PR #40, opened 2023-07-25
Rename tests directory to examples

petewall:
* Add README for each example
* Remove agent-only and all-disabled, since they're not really useful or interesting
* Remove no-allow-lists since it's covered by custom-allow-lists


skl (approved):
Nice

PR #43, opened 2023-07-26
Update dependencies and move Makefile

petewall:
* I updated the dependent helm charts (changes in charts/k8s-monitoring and inside examples/*/output.yaml)
* I moved the Makefile that was inside examples to the top level. That Makefile is more about chart readiness and less about examples, so moving it to the top makes more sense. (changes in .github, scripts)


jewbetcha (approved):
LGTM but @skl will know better :D

skl (approved):
Great job, Pete. It's actually really helpful to see the regenerated outputs here. Thanks!

PR #50, opened 2023-08-02
Support custom rules for each metrics

davhdavh:
Add configuration points for each of the metrics to be able to run rules against the collected metrics.

Also clean up some inconsistencies between the various metrics, and the missing agent metrics that is also collected but has no config


charts/k8s-monitoring/templates/agent_config/_cadvisor.river.txt
```diff
@@ -38,9 +38,10 @@ prometheus.scrape "cadvisor" {
 prometheus.relabel "cadvisor" {
   rule {
     source_labels = ["__name__"]
-    regex = "up|{{ join "|" .Values.metrics.cadvisor.allowList }}"
+    regex = "up|{{ join "|" (index .Values.metrics "cadvisor").allowList }}"
```
petewall, line 41: Why change this? We were using the `index` method for `kube-state-metrics` and `node-exporter` because they had dashes in them. We don't need to make this change here.

charts/k8s-monitoring/templates/agent_config/_agent.river.txt
```diff
@@ -33,9 +33,10 @@ prometheus.scrape "agent" {
 prometheus.relabel "agent" {
   rule {
     source_labels = ["__name__"]
-    regex = "up|agent_build_info"
+    regex = "up|{{ join "|" (index .Values.metrics "agent").allowList }}"
```
petewall, line 36: This is helpful!

petewall (changes requested):
Can we change `custom_rules` to `additionalMetricRelabelingRules`? That makes it more accurate of a name.

charts/k8s-monitoring/templates/agent_config/_cadvisor.river.txt
davhdavh: For consistency with the other files

davhdavh:
It takes forever to do the doc/unittest loop because they use some custom tools for it, I dont really want to install

petewall (approved):
Looks good to me.
Also, bump the chart version and re-run helm-docs.

petewall:
This PR was incorporated in #92. Closing this one.
Thank you, @davhdavh for the contribution. I made sure that you got Co-Author credit on #92.

PR #51, opened 2023-08-02
Add Windows Exporter as a new optional metric source.

petewall:
Fixes #29

PR #55, opened 2023-08-03
Add the missing nodeSelectors for windows

davhdavh:
Current config would only work in tainted windows nodes environment. Add nodeSelector to all subcharts to force them on linux machines


davhdavh:
Ref: #51

petewall (approved):
Looks good to me.

Can you bump the version in Chart.yaml and run helm-docs again?

skl (approved):
I merged and tested this in #62 👍

PR #58, opened 2023-08-10
Add additional windows exporter -> linux metricsnames mappings

davhdavh:
Map memory, and container level metrics as best as possible to linux equivalent

### DIFF
--- a/charts/k8s-monitoring/README.md
+++ b/charts/k8s-monitoring/README.md
@@ -272,11 +272,16 @@ Visit the Kube State Metrics [documentation](https://github.com/kubernetes/kube-
 
 ### Allow List for Windows Exporter
 
-* windows_.*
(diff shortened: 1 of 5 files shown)


skl:
You're right about the missing labels from some of the windows exporter metrics. For those metrics, a recording rule might be better suited than a relabel, with a query similar to the one on their docs [here](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.container.md#useful-queries).

A recording rule is particularly suited to the `kube_pod_info` join due to that being a high cardinality metric.

petewall:
As per #358, we're going to undo trying to make Windows exporter behave like Node exporter. It's chasing too much of a moving target, and some things don't translate correctly.
Thanks for bringing this PR, but I'm going to have to close it.

PR #59, opened 2023-08-14
feat: include 'tenantId' field in values for on-prem mimir & loki auth

hbjydev:
This MR adds the ability to push tenant IDs with the Loki & Prometheus configs in the format Mimir and Loki allow for.

This is to make the system usable by people like us who run Mimir & Loki OSS on-prem, and need to set the tenant IDs in the Grafana Agent config.

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_prometheus.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_prometheus.river.txt
@@ -15,7 +15,7 @@ local.file "prometheus_password" {
   is_secret = true
 }
 {{- end }}
-{{ if .basicAuth.password }}
+{{ if .tenantId }}
 local.file "prometheus_tenantid" {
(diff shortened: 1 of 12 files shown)


charts/k8s-monitoring/templates/agent_config/_prometheus.river.txt
```diff
@@ -15,17 +15,26 @@ local.file "prometheus_password" {
   is_secret = true
 }
 {{- end }}
+{{ if .basicAuth.password }}
```
TomasGrbalik, line 18: Hello! Just wondering, shouldn't this be concerned with tenantId? 
```suggestion
{{ if .tenantId }}
```

hbjydev:
Thanks @petewall

FYI, this currently won't support OpenCost, but I have an issue open with them to implement that functionality :) (opencost/opencost#2094)

PR #76, opened 2023-08-18
Value for opencost CLUSTER_ID

### LINKED ISSUE #48: Cluster name for opencost cost is 'default-cluster'
It seems the cluster.name config is not applied to opencost

skl:
Fixes #48

The `CLUSTER_ID` environment variable on the Opencost pod is set to a default value of `default-cluster` in the Opencost helm chart, so we provide a value to override this.


jewbetcha (approved):
LGTM 👍

PR #79, opened 2023-08-21
Adding useFullAllowList for node-exporter

gaantunes:
For Linux Node integration to work fully, the set of metrics collected by node exporter needs to be extended.

To asbtract the complexicity from the customer and allow for a smaller command to activate the right list of metrics on the helm chart, this is adding a new flag to Values.yaml, which should be used to activate the Linux Node integration.

The customer will need to run:

`helm upgrade --install grafana-k8s-monitoring  grafana/k8s-monitoring --namespace "default" --set metrics.node-exporter.service.useFullAllowList=true --reuse-values`


petewall:
Closing as this will be addressed in: https://github.com/grafana/k8s-monitoring-helm/pull/326

PR #80, opened 2023-08-22
Otel collector example removal

peterolivo:
changed username/password to default for otel example


petewall (changes requested):
This example actually was for an unfinished feature.
Can you simply remove the whole examples/otel-collector directory? It doesn't belong here at all anyway.

petewall:
Note, this access policy token does not exist anymore.

petewall (approved):

PR #92, opened 2023-08-25
Feat: Allow for extra rules and stages

petewall:
* Allow for extra relabeling and metric relabeling rules for every metric source.
* Allow for extra relabeling rules and processing stages for pod logs 
* Update example READMEs


jewbetcha (approved):
LGTM

PR #105, opened 2023-09-05
Add scrape interval to all metric sources

### LINKED ISSUE #84: Add scrape intervals to each metric?
I've heard that there was confusion over the default scrape interval being used when scraping metrics.

The default is `60s`, which is set in [prometheus.scrape](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.scrape/#arguments). However, a user not familiar with flow might not know that is the default. They wouldn't necessarily know to dig down into agent documentation.

If we set:
```
node-exporter:
  scrapeInterval: 60s
```

And use that in the generated config, it would make it explicit.

petewall:
Fixes #84

### DIFF
--- a/charts/k8s-monitoring/Chart.yaml
+++ b/charts/k8s-monitoring/Chart.yaml
@@ -3,7 +3,7 @@ name: k8s-monitoring
 description: A Helm chart for gathering, scraping, and forwarding Kubernetes infrastructure metrics and logs to a Grafana Stack.
 type: application
 
-version: 0.1.15
+version: 0.1.16
 appVersion: 1.2.0
 icon: https://raw.githubusercontent.com/grafana/grafana/main/public/img/grafana_icon.svg
 maintainers:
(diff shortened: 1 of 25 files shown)


jewbetcha (approved):
LGTM 👍 IIRC the scrape config as it exists out in the wild right now is 30s? maybe we should advise an update after this goes out

petewall:
> IIRC the scrape config as it exists out in the wild right now is 30s?

Actually, this effectively is a no-op. Grafana agent's default scrape interval is 60s (https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.scrape/#arguments), not 30s. We're just pulling that variable up a layer to make it easier to override.

jewbetcha:
Ah ok, maybe I was thinking about something else that caused our scrapes to increase. Sounds good.

rknightion:
@petewall would it be possible to run the create-release ci? - Looks like although the chart ver was updated with this PR a release wasn't cut

petewall:
Thanks for the poke. Yep, it's done now.

PR #108, opened 2023-09-06
Add github action to update dependencies

### LINKED ISSUE #106: Use Updatecli to make a "dependabot"-like experience for Chart dependencies
I'd love for GitHub's dependabot to be able to make PRs for helm chart dependencies, but it doesn't look like that's coming any time soon.

Instead, you can run [updatecli](https://www.updatecli.io/docs/automate/github_action/) in a GitHub action that periodically checks and creates PRs.

petewall:
Fixes #106 

This action generated the `[dependency] ...` Pull Requests that are open right now.

Note for reviewers:
* Updatecli (if you're not familiar) knows how to check helm chart repositories for updates and will locally modify our helm chart source with the updated version. It seems to have the ability to create PRs, but we need to run `helm-docs` and regenerate the examples before making the PR.  Also, their docs on PR creation are outdated and I couldn't get it to work reliably.
* Each .github/configs/updatecli.d/* and their associated section in the workflow are nearly identical. Review for clarity of a single flow, then check for copy-paste errors.

### DIFF
--- a/.github/workflows/check-for-dependency-updates.yaml
+++ b/.github/workflows/check-for-dependency-updates.yaml
@@ -1,0 +1,233 @@
+---
+name: Check for dependency updates
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Run once a day
+    - cron: '0 0 * * *'
+
+permissions:
+  contents: "write"
+  pull-requests: "write"
+
+env:
+  UPDATECLI_CONFIG_DIR: "${{ github.workspace }}/.github/configs/updatecli.d"
+
+jobs:
+  updateGrafanaAgent:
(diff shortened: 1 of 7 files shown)


.github/workflows/check-for-dependency-updates.yaml
```diff
@@ -0,0 +1,281 @@
+---
+name: Check for dependency updates
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Run once a day
+    - cron: '0 0 * * *'
```
jewbetcha, line 8: This is nice, didn't realize you could make a cron job in a workflow

skl (approved):
lgmt, thank you!

PR #120, opened 2023-09-11
Set signoff: true for dependency update commits

petewall:


skl (approved):

petewall:
Actually, this will not solve the issue. We need signed commits.
https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#gpg-commit-signature-verification

Issue #101, opened 2023-08-31
Tracking issue for `loki.source.kubernetes`


skl:
Since #61 we're no longer using `loki.source.kubernetes`, this issue tracks the possibility of using it again in future by monitoring related issues:

- https://github.com/grafana/agent/issues/4502
- https://github.com/grafana/agent/issues/4098
- https://github.com/grafana/agent/issues/5040
  - https://github.com/kubernetes/kubernetes/issues/59902
  - https://github.com/kubernetes/kubernetes/pull/118500

petewall:
This same issue also applies to [loki.source.podlogs](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.podlogs/), so we should hold off on supporting PodLogs until the underlying issue is resolved.

If someone really needs PodLogs, it should be pretty simple to add it in with `.logs.extraConfig`

petewall:
Perhaps we can re-enable API-delivered logs as an option.
Defaulting to DaemonSet and hostPath mounts, but have an option in the values to use the API method from the statefulset agent. This would allow for logs in clusters like Fargate, or other places where mounting host-path volumes would be prohibited.

The agent is [adding code](https://github.com/grafana/agent/pull/5623) that should address the limitation.

petewall:
I'm going to close this. We're definitely allowing the use of API based log gathering. If that method becomes the norm, we can adjust the config as needed.

PR #124, opened 2023-09-12
Add support for Probes and add namespace selectors for each object type

petewall:
Skipping PodLogs, because it has the same issue as #101 and also must be loaded by a single agent instance. See the [loki.source.podlogs docs](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.podlogs/) for more information.


charts/k8s-monitoring/templates/agent_config/_grafana_operator_objects.txt
```diff
@@ -0,0 +1,79 @@
+{{ define "agent.config.pod_log_objects" }}
+// PodMonitor
+loki.source.podlogs "pod_logs" {
```
skl, line 3: This isn't referenced again in this file, is it still needed?

skl (approved):
lgtm - when you say probes, you mean like [these kinds of probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request)?

charts/k8s-monitoring/templates/agent_config/_grafana_operator_objects.txt
petewall: Oop, yes. Thanks. Removed that.

PR #128, opened 2023-09-14
Add agent metrics configuration and scrape all logs/metrics agent pods

n888:


charts/k8s-monitoring/templates/agent_config/_agent.river.txt
```diff
@@ -1,41 +1,77 @@
...
     enabled = true
   }
 {{- end }}
+{{- if .Values.metrics.agent.pod.isTLS }}
+  scheme = "https"
+  bearer_token_file = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+  tls_config {
+    insecure_skip_verify = true
+  }
+{{- end }}
```
petewall, line 58: Remove this. We haven't needed TLS for the agent, so let's not complicated it until its required.

charts/k8s-monitoring/values.yaml
```diff
@@ -82,6 +82,82 @@ metrics:
...
+    # -- (string) Rule blocks to be added to the prometheus.relabel component for Grafana Agent.
+    # See https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.relabel/#rule-block
+    extraMetricRelabelingRules:
+
+    # Grafana Agent pod settings
+    pod:
+      # -- Name of the metrics port
+      port: http-metrics
+      # -- Does this port use TLS?
+      isTLS: false
```
petewall, line 112: Same as in the river, just remove this and hardcode the `http-metrics` port name. This chart controls the deployment of the Grafana Agent, so we don't need ultimate configurability.

charts/k8s-monitoring/values.yaml
```diff
@@ -82,6 +82,82 @@ metrics:
...
+    - prometheus_remote_storage_bytes_total
+    - prometheus_remote_storage_highest_timestamp_in_seconds
+    - prometheus_remote_storage_samples_failed_total
+    - prometheus_remote_storage_samples_retried_total
+    - prometheus_remote_storage_samples_total
+    - prometheus_remote_storage_sent_batch_duration_seconds_bucket
+    - prometheus_remote_storage_sent_batch_duration_seconds_sum
+    - prometheus_remote_storage_shards
+    - prometheus_remote_storage_shards_max
+    - prometheus_remote_storage_shards_min
```
petewall, line 159: We gotta really pare this down. The allowList should be a pretty minimal set, unless there's really good reason.
I'd keep it to just `agent_build_info`. I'll look at the rest of the list and see if there's a compelling reason to keep more.

charts/k8s-monitoring/templates/agent_config/_agent.river.txt
```diff
@@ -1,41 +1,77 @@
...
+    source_labels = ["__meta_kubernetes_pod_label_{{ $k | replace "." "_" | replace "/" "_" }}"]
+    regex = "{{ $v }}"
     action = "keep"
   }
+  {{- end }}
   rule {
-    source_labels = ["__meta_kubernetes_service_port_name"]
-    regex = "http-metrics"
+    source_labels = ["__meta_kubernetes_pod_container_port_name"]
+    regex = "{{- .Values.metrics.agent.pod.port -}}"
```
petewall, line 19: Change this back to `regex = "http-metrics"`
I'll explain down below.

petewall (commented):
This is really useful. Thanks!

A few changes: I think it's good to scrape more than just the metrics pods, we should be careful about adding too much functionality and customizability before it's really needed. Also, I want to be careful about adding too many more metrics to a default set.

n888:
@petewall Thanks for the pointers! I have added a blurb for Agent Metrics list in the readme as well (not sure if those links are the best metric list sources)

Please review, cheers

petewall (changes requested):
This is looking great! One final change, the `README.md` file is actually generated by [helm-docs](https://github.com/norwoodj/helm-docs) from the `README.md.gotmpl` file. Can you move the "### Allow List of Grafana Agent" section to the template file, then run `helm-docs`. That will automatically bring in the changes to the values (it generates that section from reading the values.yaml.

n888:
All set, thanks !

petewall:
One last change required:
The examples are actually generated from the real chart. So, at the top level of the repository, can you run `make regenerate-example-outputs` and add those changes? That should fix the failed checks.

n888:
Gotcha -- Pushed :+1:

petewall (approved):

petewall:
Another PR just came in. Can you rebase against the latest from main? Much of the content from `grafana-agent-config.yaml` has gone into `_helpers.tpl`. `make regenerate-example-outputs` might be needed again, just FYI.

petewall:
We'll get this merged!

n888:
:+1:
Should be all set, make regen didnt produce any differences

petewall:
Thanks!
I'm going to do some testing and I'll try and cut a release tomorrow.

PR #132, opened 2023-09-14
Add an example of filtering to a specific namespace.

### LINKED ISSUE #121: Add doc/example for restricting to a specific namespace
Add example to show how to restrict metrics, logs, and events to a specific namespace. Be clear that you can mix and match each of these.

petewall:
Also turn toStrings to toJson, which does what we want in the river format.

Fixes #121


charts/k8s-monitoring/values.yaml
```diff
@@ -389,6 +389,9 @@ logs:
     # -- Capture and forward logs from Kubernetes pods
     enabled: true
 
+    # -- Only capture logs from pods in these namespaces (empty means all namespaces)
```
skl, line 392:
```suggestion
    # -- Only capture logs from pods in these namespaces (`[]` means all namespaces)
```

Wonder if this is worth clarifying? and if helm docs supports markdown formatting... same for line 413.

charts/k8s-monitoring/templates/agent_config/_cluster_events.river.txt
```diff
@@ -1,7 +1,12 @@
 {{ define "agent.config.logs.cluster_events" }}
+{{- with .Values.logs.cluster_events }}
 // Cluster Events
 loki.source.kubernetes_events "cluster_events" {
   job_name   = "integrations/kubernetes/eventhandler"
+{{- if .namespaces }}
```
skl, line 6: This syntax `.namespaces` is it referring to a top-level value (which I couldn't find) or is it relative? Wondering why this is so short compared to `.Values.logs.cluster_events` above.

skl (approved):
nice work

charts/k8s-monitoring/templates/agent_config/_cluster_events.river.txt
petewall: Ah, this is a helm'ism. Or at least a go template idiom. Earlier, there's a `{{- with .Values.logs.cluster_events }}` which changes the context of `.` to be at the cluster events level. It's just useful for scoping things and not having to fully qualify every time.

skl: Ah ok, thanks for explaing that 👍

skl (approved):
lgtm

PR #139, opened 2023-09-19
Update Chart.yaml add sourse

Vad1mo:
Add source, to point to the charts source code. 
Very helpful if the chart is displayed based on chart.yaml file and there is no reference where the code is.

In Particular in artifact hub. it is helpful, to know where the templates are.
 
Bildschirmfoto 2023-09-19 um 15 31 28


petewall (approved):

petewall:
This is great! Thanks for adding this.
This change requires an update to the README.md.  The easiest way to do that is to run [helm-docs](https://github.com/norwoodj/helm-docs) in the chart directory and commit the README.md file.

PR #140, opened 2023-09-19
Adding extra attributes/actions capabilities to traces

marioanton:
Adding extra attributes/actions capabilities to traces


petewall:
This is a great idea. Can you add `extraAttributes` to the traces section in values.yaml? Maybe augment the traces example with something to help other people know how to make use of this feature?

petewall:
I'd also be curious if anything changes if/when this feature is added to the Agent: https://github.com/grafana/agent/pull/5229

marioanton:
I understand that wont be affected since k8sattributes field will only process metadata coming from k8s infra (pods).
Might need some extra logic for that big when enabled to join the config towards an easier way to compound the final attributes config.

PR #141, opened 2023-09-19
Add a README for the examples directory and improve some other docs

petewall:


skl (approved):
Awesome, might be worth adding to the "not generated" list in gitattributes

PR #145, opened 2023-09-22
Add HTTP port to trace ingestion

petewall:


charts/k8s-monitoring/templates/_helpers.tpl
```diff
@@ -1,22 +1,22 @@
 {{/* This template checks that the port defined in .Values.traces.receiver.port is in the targetPort list on .grafana-agent */}}
```
skl, line 1:
```suggestion
{{/* This template checks that the port defined in .Values.traces.receiver.*.port is in the targetPort list on .grafana-agent */}}
```

charts/k8s-monitoring/templates/agent_config/_traces.river.txt
```diff
@@ -1,16 +1,26 @@
 {{ define "agent.config.traces" }}
-{{- include "checkForTracePort" . -}}
-{{- with .Values.traces }}
 // Traces
 otelcol.receiver.otlp "trace_receiver" {
+{{- if .Values.traces.receiver.grpc.enabled }}
```
skl, line 4: The older .receiver.port was kept for compatibility but looks like it was dropped from the template iiuc - do you need another conditional to check for that?

skl (approved):
lgtm - just a compatibility comment

charts/k8s-monitoring/templates/agent_config/_traces.river.txt
petewall: Yes, thank you!

PR #150, opened 2023-09-25
Allow for external secrets

petewall:


petewall:
Unanswered questions:
1. What all needs to go into the secret?
  * host
  * username
  * password
  * tenant id
  * Write endpoint?
  * proxyURL?
  * externallabels?
2. How do we build the configuration if we don't know the authentication options?
For example: Currently, we build a component like this if the username and password are both provided:
```
    prometheus.remote_write "grafana_cloud_prometheus" {
      endpoint {
        url = <host> + "/api/prom/push"
        basic_auth {
          username = <username>
          password = <password>
        }
      }
```
but, if the password is not supplied in the chart, we don't have the password component.

PR #154, opened 2023-09-27
Add a helm hooks and tests to validate config and run queries

petewall:
This change does two things:
First, on installs and upgrades, before anything is actually deployed, we deploy a pod running the agent image and have it run `agent fmt` against the configuration (duplicated for logs config). This will check for syntax formatting errors early and bail on the deployment.

Second, it adds a test job, which runs when you run `helm test` *after* the deployment is done. This spins up a pod and runs curl commands against the prometheus service to check that the metrics actually have arrived.

NOTE: Some limitations of the test:
* It doesn't check for logs or events with Loki... that's for a future PR
* It also doesn't time-box the queries, which probably would make sense. Run the query so it only returns since the last deployment, otherwise, it could get very old data.


skl:
@petewall this looks awesome, I'd be excited to see the query tests running but I can't see `Running query` in the log output of the test chart action against the commit [[link](https://github.com/grafana/k8s-monitoring-helm/actions/runs/6408849934/job/17398751923)]. Should I expect to see the `up` queries running there?

petewall:
It looks like the `helm test` does not print logs from the test pod unless `--logs` is passed to it.
Let me check if I can configure `ct` to pass that along

petewall:
It might not be possible, unfortunately, since there isn't a way to pass args to just the `helm test` portion of chart-testing:
https://github.com/helm/chart-testing/issues/540

skl (approved):
Thanks for this!

PR #157, opened 2023-10-02
Add static telemetry metric with informaiton about chart version and enabled features

### LINKED ISSUE #143: Create some sort of static metric that reports chart version and some values
It would be great to be able to provide some sort of static metric like:
```
kubernetes_monitoring_build_info{
  version="0.2.4",
  metrics="agent,node-exporter,kube-state-metrics,kubelet,cadvisor",
  logs="true",
  events="true",
  traces="false"
} = 1
```

Something that reports up to prometheus if the chart was used to deploy, the version of the chart, and a summary of the settings used.

petewall:
This creates a config map with a static metric defined: `grafana_kubernetes_monitoring_build_info`.
That metric has five labels:
* `version` - the chart version
* `metrics` - If metrics are enabled, will list the metric sources that are also enabled: "agent", "kube-state-metrics", "node-exporter", "windows-exporter", "kubelet", "cadvisor", "cost"
* `logs` - If logs are enabled, will list "events" and/or "pod_logs"
* `traces` - "enabled" or "disabled" based on the state of `.Values.traces.enabled`
* `deployments` - The list of subcharts that are being deployed

The configmap is mounted to the agent pod, then scraped using `prometheus.exporter.unix`'s textfile scraper.

Fixes #143

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_kubernetes_monitoring_telemetry.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_kubernetes_monitoring_telemetry.river.txt
@@ -1,0 +1,28 @@
+{{ define "agent.config.kubernetes_monitoring_telemetry" }}
+// Kubernetes Monitoring Telemetry
+prometheus.exporter.unix {
+  set_collectors = ["textfile"]
+  textfile {
+    directory = "/etc/kubernetes-monitoring-telemetry"
+  }
+}
+
+prometheus.scrape "kubernetes_monitoring_telemetry" {
+  job_name   = "integrations/kubernetes/kubernetes_monitoring_telemetry"
+  targets    = prometheus.exporter.unix.targets
+  scrape_interval = {{ .Values.metrics.scrapeInterval | quote }}
+{{- if (index .Values "grafana-agent").agent.clustering.enabled }}
+  clustering {
+    enabled = true
+  }
+{{- end }}
+  forward_to = [prometheus.relabel.kubernetes_monitoring_telemetry.receiver]
+}
+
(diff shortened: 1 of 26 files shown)


charts/k8s-monitoring/templates/agent_config/_kubernetes_monitoring_telemetry.river.txt
```diff
@@ -0,0 +1,28 @@
...
+  clustering {
+    enabled = true
+  }
+{{- end }}
+  forward_to = [prometheus.relabel.kubernetes_monitoring_telemetry.receiver]
+}
+
+prometheus.relabel "kubernetes_monitoring_telemetry" {
+{{- if .Values.metrics.extraMetricRelabelingRules }}
+{{ .Values.metrics.extraMetricRelabelingRules | indent 2 }}
```
skl, line 24: Does it still make sense to allow relabelling here for `kubernetes_monitoring_telemetry`? I'm not sure of the use case for relabelling that metric, I can only think of ways it could break or be abused... 😅

skl (approved):
This is awesome, thank you!

charts/k8s-monitoring/templates/agent_config/_kubernetes_monitoring_telemetry.river.txt
petewall: My thinking: This allows people to add custom labels if they wanted (e.g. `region="netherton"`)

skl: Ok, I see. Makes sense then 👍

PR #168, opened 2023-10-10
Fixes for OpenShift

### LINKED ISSUE #164: Further improvements for OpenShift
Just found an OpenShift deployment that was not working with logs.

It had issues scheduling the logs daemonset because of hostPath and other security context violations.

Adding:
```
      securityContext: # CHANGED
        seLinuxOptions:
          type: spc_t
```
Would probably solve it.

petewall:
Fixes #164

### DIFF
--- a/charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
+++ b/charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
@@ -19,7 +19,6 @@ fsGroup:
   type: RunAsAny
 groups: []
 priority: null
-readOnlyRootFilesystem: true
 requiredDropCapabilities:
 - CHOWN
 - DAC_OVERRIDE
--- a/charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
+++ b/charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
@@ -1,0 +1,48 @@
+{{- if and (eq .Values.cluster.platform "openshift") .Values.logs.enabled .Values.logs.pod_logs.enabled }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{ include "grafana-agent.fullname" (index .Subcharts "grafana-agent-logs") }}
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+allowedCapabilities: null
+defaultAddCapabilities: null
+defaultAllowPrivilegeEscalation: false
+forbiddenSysctls:
+- '*'
+fsGroup:
+  type: RunAsAny
+groups: []
+priority: null
+readOnlyRootFilesystem: true
+requiredDropCapabilities:
+- CHOWN
+- DAC_OVERRIDE
+- FSETID
+- FOWNER
+- SETGID
+- SETUID
+- SETPCAP
+- NET_BIND_SERVICE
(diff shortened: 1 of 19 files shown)


examples/openshift-compatible/README.md
```diff
@@ -45,8 +46,25 @@ prometheus-node-exporter:
 grafana-agent:
   agent:
     listenPort: 8080
+    securityContext:
```
theSuess, line 49: Is there an argument against applying this securityContext to all platforms? I'm all for safer defaults :wink:

charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if and (eq .Values.cluster.platform "openshift") .Values.logs.enabled .Values.logs.pod_logs.enabled }}
```
theSuess, line 1: When comparing with the [openshift-logging scc](https://github.com/openshift/cluster-logging-operator/blob/master/internal/auth/securitycontextconstraint.go#L48), we're missing the `readOnlyFileSystem` option. We should be able to enable this

charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
```diff
@@ -0,0 +1,47 @@
...
+  type: RunAsAny
+seccompProfiles:
+  - runtime/default
+supplementalGroups:
+  type: RunAsAny
+users: []
+volumes:
+- configMap
+- emptyDir
+- hostPath
```
theSuess, line 44: This is not present in the openshift logging scc, maybe we don't need this?

charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
petewall: I had to remove that because the Agent fails when it tries to write to `/tmp/agent`.
I could add that back in and mount an `emptyDir` volume for temp storage

charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
petewall: I think it is there, right?
```
var DesiredSCCVolumes = []security.FSType{"configMap", "secret", "emptyDir", "projected"}

func NewSCC() *security.SecurityContextConstraints {
...
	scc.Volumes = DesiredSCCVolumes
```

We're adding `hostPath` so the agent can read logs.

petewall: https://github.com/openshift/cluster-logging-operator/blob/master/internal/auth/securitycontextconstraint.go#L30C1-L39

theSuess: I was referencing the `hostPath` value. I'm wondering how openshift logging accomplishes the same without it.

Also don't see this as a blocker. The changes look good as is, I'm just checking if we can do with less permissions :

charts/k8s-monitoring/templates/platform_specific/openshift/security-context-constraint.yaml
theSuess: Alright, I figured out how the openshift log collector does it.

They store log position information in `/var/lib/vector` and as this is a mounted volume (not the root volume) they can set keep the root volume to read only.

We could try it with `/var/lib/grafana-agent`

examples/openshift-compatible/README.md
petewall: it's a good discussion that we should have with the Agent team.

PR #174, opened 2023-10-12
Add schema file

petewall:


petewall:
This schema is "tested" by regenerating the examples.

skl (approved):
JSON Schema, there's something you don't see every day. How do we update the schema when new values are introduced?

PR #175, opened 2023-10-13
fix readme

cornelius-keller:
While deploying to my cluster I noticed that the example values.yaml in the readme is not correct. 
The `username` and `password` values should be under the key `basicAuth` for both loki and prometheus.


petewall:
Ooh! Thank you so much for this. Good catch. Can you make the same change in `charts/k8s-monitoring/README.md.gotmpl`? That's the source file for `README.md` and we use [helm-docs](https://github.com/norwoodj/helm-docs) to generate it.

petewall (approved):

PR #176, opened 2023-10-13
Only add the RBAC, rather than replacing them

petewall:
Actually, RBAC rules are additive, so we can simply add this RBAC rule and let the agent create their own again


skl (approved):
🚀

PR #179, opened 2023-10-17
Add a check for the grafana agent crd

petewall:


petewall:
Tested on kind:
1. Install on fresh cluster works
2. Upgrades work
3. Install after deploying standalone grafana-agent chart fails with error message. Adding `grafana-agent.crds.create = false` worked.

captncraig:
I wish this wasn't needed, but it is probably the cleanest solution. 

I've hit up against it, and my solution was to delete the offending crds, then run this.

My only question is: should this be made in the agent chart instead?

captncraig (approved):

petewall:
I mean, that's up to your squad. It's really not preventing the error, just adding a friendly error message and some guidance.
In this chart, it's a bit "special", since I'm deploying a lot of CRDs (agent and prom operator), but I don't control them.

PR #191, opened 2023-10-19
Add tenantID support for tempo

bytheway:


petewall:
I like this change, but there are a few extra steps to ensure that the generated content stays up to date.
Can you run the `helm-docs` and `make regenerate-example-outputs` steps. See https://github.com/grafana/k8s-monitoring-helm#contributing for more context.

bytheway:
Can do, I should've checked that doc before throwing these over the fence.  I'll get the additional work done later today. Thanks!

petewall (approved):

PR #197, opened 2023-10-20
Add the ability to specify extra queries for the test job

petewall:


.github/configs/loki.yaml
```diff
@@ -19,3 +19,5 @@ monitoring:
     enabled: false
     grafanaAgent:
       installOperator: false
+  serviceMonitor:
+    enabled: false
```
petewall, line 23: This is unrelated, but removes accidental extra metrics.

skl (approved):
What does a failure look like? e.g. is an empty response a failure, or just bad PromQL?

I mean, what difference does it make which query you test here, is there some kind of implied pass criteria?

petewall:
> What does a failure look like? e.g. is an empty response a failure, or just bad PromQL?
> 
> I mean, what difference does it make which query you test here, is there some kind of implied pass criteria?

A failure will print the failure reason. Either a failure to send the query, or an empty result. So, a bad PromQL query is one failure mode, a valid PromQL query with no results is also a failure mode. The test job will run a number of times to allow for the time to start, scape, and export the metrics.

PR #211, opened 2023-10-25
For pod logs, filter to the pods only on the same node

petewall:
Limit the amount of pods discovered by filtering to only the pods on the same node as the agent. This reduces the amount of pods that need to be filtered out later on.

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_pod_logs.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_pod_logs.river.txt
@@ -1,8 +1,14 @@
 {{ define "agent.config.logs.pod_logs" }}
 {{- with .Values.logs.pod_logs }}
 // Pod Logs
+discovery.kubelet "pods" {
+{{- if .namespaces }}
+  namespaces = {{ .namespaces | toStrings | toJson }}
+{{- end }}
+}
+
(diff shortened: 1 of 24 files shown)


charts/k8s-monitoring/templates/agent_config/_pod_logs.river.txt
```diff
@@ -1,9 +1,23 @@
 {{ define "agent.config.logs.pod_logs" }}
 {{- with .Values.logs.pod_logs }}
 // Pod Logs
+discovery.kubernetes "pods" {
+    role = "pod"
```
skl, line 5: Couple of questions about this line. One is a nit: the indentation isn't quite right. The other is my real question: is it intentional for this line to appear both here and nested under `selectors` below?

skl (approved):

charts/k8s-monitoring/templates/agent_config/_pod_logs.river.txt
petewall: Yeah, it's odd behavior, but it's correct: https://grafana.com/docs/agent/latest/flow/reference/components/discovery.kubernetes/#limit-to-only-pods-on-the-same-node

PR #213, opened 2023-10-26
Make an option for the metrics service to use otlp protocol instead of prometheus remote write

petewall:


rfratto (approved):
LGTM, but I do think the three protocols should be remote_write/otlp/otlphttp since OTLP is the name of the protocol and not otel :)

eamonryan (approved):
LGTM :+1:

petewall:
Tests still failing. It's deploying correctly, it just the queries aren't coming back. I'll look at it in the morning.

petewall:
Found the issue, the agent can't send the metrics to prometheus as configured:
```
ts=2023-10-26T22:11:31.869089413Z level=info msg="Exporting failed. Will retry the request after interval." component=otelcol.exporter.otlp.metrics_service error="rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: first record does not look like a TLS handshake\"" interval=28.327073829s
```

Need to fiddle with it to fix the test environment.

PR #218, opened 2023-10-30
Add support for otel metrics push with prometheus and otlphttp

tekicat:
This PR provides the ability to publish metrics collected via opentelemetry instrumentation to prometheus via promethues rewrite or otelcol.exporter.otlphttp or to both simultaneously


cedricziel:
I think the change in https://github.com/grafana/k8s-monitoring-helm/pull/219 needs to be applied here as well in order for the metrics to have the correct cluster name resource attribute

petewall:
I'm closing this, because the `prometheus` service (which really should just be called "metrics service") has the option to use otlp or otlphttp protocols when pushing metrics.

PR #219, opened 2023-10-31
Add k8s.cluster.name through resource attributes

cedricziel:
The concrete change here is injecting a new transform processor because the attributes processor cannot write to/access the resource.

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_traces.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_traces.river.txt
@@ -43,7 +43,20 @@ otelcol.processor.batch "trace_batch_processor" {
   send_batch_max_size = {{ .processors.batch.maxSize | int }}
   timeout = {{ .processors.batch.timeout | quote}}
   output {
-    traces = [otelcol.processor.attributes.trace_attributes_processor.input]
+    traces = [otelcol.processor.attributes.trace_transform_processor.input]
+  }
+}
+
+otelcol.processor.transform "trace_transform_processor" {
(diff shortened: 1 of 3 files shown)


charts/k8s-monitoring/templates/agent_config/_traces.river.txt
```diff
@@ -43,22 +43,31 @@ otelcol.processor.batch "trace_batch_processor" {
...
 }
+
 {{- end }}
 
+otelcol.processor.transform "trace_transform_processor" {
+  trace_statements {
+    context = "resource"
+    statements = [
+      // Accessing a map with a key that does not exist will return nil.
+      "set(attributes[\"k8s.cluster.name\"], \"{{ required ".Values.cluster.name is a required value. Please set it and try again." .Values.cluster.name }}\") where attributes[\"k8s.cluster.name\"] == nil"
```
petewall, line 57: I think this needs a `,` at the end.

petewall (commented):
Can you help me understand what the difference is here?

cedricziel:
> Can you help me understand what the difference is here?

Absolutely @petewall . The attributes processor works on signal level attributes, while the transform processor can also set resource-level attributes. The resource level is one level above signals and the same resource would ideally be associated with concrete metrics, logs and traces.

https://opentelemetry.io/docs/specs/otel/resource/sdk/

The concrete change here is injecting a new transform processor because the attributes processor cannot write to/access the resource.

petewall:
Closing per @mar4uk 's comment.

PR #223, opened 2023-11-01
Improve the Helm test job so it can run LogQL queries

petewall:
This PR extracts the script that does the testing into its own file. That let's people run it on the CLI if they want.
It also adds a `type` field that lets the tests run PromQL or LogQL tests.

I'm not adding the LogQL queries for cluster events or pod logs to the default set of queries, though, because that requires the access token to be able to read logs from loki. By default, our access token only has write permissions to grafana cloud loki.

### DIFF
--- a/charts/k8s-monitoring/templates/tests/test.yaml
+++ b/charts/k8s-monitoring/templates/tests/test.yaml
@@ -74,7 +74,7 @@ spec:
       containers:
         - name: test-runner
           image: {{ .Values.test.image.registry }}/{{ .Values.test.image.image }}:{{ .Values.test.image.tag }}
-          command: ["bash", "-c", "apt-get update && apt-get install -y -y curl jq && bash /etc/test/test.sh /etc/test/queries.json"]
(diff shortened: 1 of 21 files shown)


charts/k8s-monitoring/templates/tests/test.yaml
```diff
@@ -91,7 +74,7 @@ spec:
       containers:
         - name: test-runner
           image: {{ .Values.test.image.registry }}/{{ .Values.test.image.image }}:{{ .Values.test.image.tag }}
-          command: ["bash", "-c", "bash /etc/test/test.sh /etc/test/queries.json"]
+          command: ["bash", "-c", "apt-get update && apt-get install -y -y curl jq && bash /etc/test/test.sh /etc/test/queries.json"]
```
skl, line 77:
```suggestion
          command: ["bash", "-c", "apt-get update && apt-get install -y curl jq && bash /etc/test/test.sh /etc/test/queries.json"]
```

skl (approved):
Nice

charts/k8s-monitoring/templates/tests/test.yaml
petewall: That's apt-get for "pretty please!"

PR #227, opened 2023-11-02
move the namespace filter after rename the kube label

dmouse:
While attempting to enable logs for specific namespaces, I observed that logs ceased being sent to Grafana Cloud. The configuration I used is as follows:

```
logs:
  cluster_events:
    enabled: true
    namespaces:
      - fakku-app
  pod_logs:
    enabled: true
    namespaces:
      - apps
```

This configuration resulted in the generation of:

```
discovery.relabel "pod_logs" {
  targets = discovery.kubernetes.pods.targets

  rule {
    source_labels = ["namespace"]
    regex = "apps"
    action = "keep"
  }
  rule {
    source_labels = ["__meta_kubernetes_namespace"]
    action = "replace"
    target_label = "namespace"
  }
```

With this configuration, I observed that no new logs were visible on my Grafana dashboard. However, after manually modifying the rules, I was able to restore log forwarding to Grafana Cloud. This pull request addresses the aforementioned issue to ensure that namespace-specific log forwarding to Grafana Cloud works as expected when you configure the namespaces in the log section.

### DIFF
--- a/examples/control-plane-metrics/output.yaml
+++ b/examples/control-plane-metrics/output.yaml
@@ -557,7 +557,6 @@ data:
     
     discovery.relabel "pod_logs" {
       targets = discovery.kubernetes.pods.targets
-    
       rule {
         source_labels = ["__meta_kubernetes_namespace"]
         action = "replace"
@@ -43489,7 +43488,6 @@ data:
     
     discovery.relabel "pod_logs" {
       targets = discovery.kubernetes.pods.targets
-    
       rule {
         source_labels = ["__meta_kubernetes_namespace"]
         action = "replace"
--- a/examples/control-plane-metrics/output.yaml
+++ b/examples/control-plane-metrics/output.yaml
@@ -563,6 +563,7 @@ data:
         action = "replace"
         target_label = "namespace"
       }
+    
       rule {
         source_labels = ["__meta_kubernetes_pod_name"]
         action = "replace"
@@ -42548,7 +42549,6 @@ metadata:
     app.kubernetes.io/version: "v0.37.3"
     app.kubernetes.io/managed-by: Helm
 spec:
-  minReadySeconds: 10
   selector:
     matchLabels:
       app.kubernetes.io/name: grafana-agent-logs
@@ -42911,7 +42911,6 @@ metadata:
 spec:
   replicas: 1
   podManagementPolicy: Parallel
-  minReadySeconds: 10
   serviceName: k8smon-grafana-agent
   selector:
     matchLabels:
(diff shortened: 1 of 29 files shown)


petewall (changes requested):
Thank you! Good catch.
Can you move the namespaces rule immediately after the rule that sets the namespace label? So, make it be the second rule. I just learned that `action = "keep"` and `action = "drop"` rules will short-circuit the rest of the evaluation, so it's worth to put it earlier in the rule chain.

Also, at the top-level of the repository, run `make regenerate-example-outputs`. That will rebuild the examples that also includes the example that uses namespaces.

examples/control-plane-metrics/output.yaml
```diff
@@ -42548,7 +42548,6 @@ metadata:
     app.kubernetes.io/version: "v0.37.3"
     app.kubernetes.io/managed-by: Helm
 spec:
-  minReadySeconds: 10
```
dmouse, line 42551: This looks like a different change that is not part of this PR, Do I need to revert this?

petewall: Hmm. yeah, that's really strange why that's being removed. That's also why the auto tests are failing below.

petewall: Ah, the Grafana Agent has this in their code:
```
spec:
  {{- if ge (int .Capabilities.KubeVersion.Minor) 22 }}
  minReadySeconds: 10
  {{- end }}
```
Since we're doing a template, it's not checking the Kubernetes version of a cluster, but likely your kubectl?
Can you run `kubectl version` and `helm version` and paste the results?

dmouse: Kubectl
```
Client Version: v1.26.9
Kustomize Version: v4.5.7
Server Version: v1.27.6
```

Helm
```
version.BuildInfo{Version:"v3.11", GitCommit:"", GitTreeState:"", GoVersion:"go1.20.1"}
```

petewall: That's so strange...
Can you edit the Makefile and change lines 25 and 26 to:
```Makefile
%/output.yaml: %/values.yaml
	helm template --kube-version 1.27.0 k8smon charts/k8s-monitoring -f $< > $@
```

Then re-run `make regenerate-example-outputs`. Check in the examples directory, but don't check-in the Makefile.

dmouse: Do I need to include this change in the Makefile?

dmouse: BTW that change works, now the `minReadySeconds: 10` is back

petewall: Great. Let's not commit the Makefile change for now. I'll keep it in mind if I want to pin that version (and have one more thing to update occasionally).

petewall (approved):

petewall:
All looks good! Thanks for the change!

PR #228, opened 2023-11-02
Remove duplicate KSM metric

ilian:
`kube_pod_status_phase` was defined twice in the allowlist of kube-state-metrics


petewall:
Thanks for this!
Can you run:
`make regenerate-example-outputs` and `cd charts/k8s-monitoring/; helm-docs`? That'll fix the errors in the tests.

petewall (approved):

PR #233, opened 2023-11-03
Add metrics for Kube Controller Manager, Kube Proxy, and Kube Scheduler

petewall:


charts/k8s-monitoring/values.yaml
```diff
@@ -403,6 +403,78 @@ metrics:
...
+
+    # -- The list of Kube Controller Manager metrics that will be scraped by the Agent
+    allowList: []
+
+  # Metrics from the Kube Proxy
+  kubeProxy:
+    # -- Scrape metrics from the Kube Proxy
+    enabled: false
+
+    # -- (int) Port number used by the Kube Proxy, set by --secure-port.
```
skl, line 435: I think kube-proxy doesn't have `--secure-port`, it has `--metrics-bind-address` and it is HTTP only (not HTTPS). Also the value for this arg needs to contain both the IP and the port like `0.0.0.0:10249`.

Here's [the doc](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/) with the CLI args and I saw scrape failures when I tried https (as the scheme was not documented). It's inconsistent with controller manager and scheduler for some reason.

Looks like you have the correct config in `.github/configs/cluster-config.yaml` but not sure if you need to specify the http scheme maybe in the scrape component.

charts/k8s-monitoring/templates/agent_config/_kube_proxy.river.txt
```diff
@@ -0,0 +1,56 @@
...
+{{ .Values.metrics.extraRelabelingRules | indent 2 }}
+{{- end }}
+{{- if .Values.metrics.kubeProxy.extraRelabelingRules }}
+{{ .Values.metrics.kubeProxy.extraRelabelingRules | indent 2 }}
+{{- end }}
+}
+
+prometheus.scrape "kube_proxy" {
+  job_name   = "integrations/kubernetes/kube-proxy"
+  targets    = discovery.relabel.kube_proxy.output
```
skl, line 30:
```suggestion
  targets    = discovery.relabel.kube_proxy.output
  scheme     = "http"
```

.github/configs/cluster-config.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kind.x-k8s.io/v1alpha4
+kind: Cluster
+nodes:
+  - role: control-plane
+    kubeadmConfigPatches:
```
skl, line 5: In my testing I found that, in addition to the configmap patches like you have here, I also had to modify the manifests to reflect these changes and add the port config too, e.g. for controller manager `/etc/kubernetes/manifests/kube-controller-manager.yaml`:

```
ports:
- containerPort: 10257
  name: https-metrics
  protocol: TCP
```

Not sure if this `kubeadmConfigPatches` handles that already?

skl (commented):
Nice work! Would like to see an example of this up and running.

charts/k8s-monitoring/values.yaml
petewall: Scraping is done by http by default. the `prometheus.scrape "kube_proxy"` component does not set https like the other ones do. But I should fix the comment.

skl (approved):

PR #236, opened 2023-11-07
Modern k8s integrations

petewall:
This PR enables new methods for adding new metric and log sources to the Kubernetes Monitoring deployment.
Specifically, it enables scraping pods and services based on annotations alone.
It also shows an example of using this method and for using an Agent module for loading integrations.

Notes to reviewers:
Maybe best to start with the changes to [values.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-cd22d083c6f49784d198f55b990feb9ae245abf1353dd56a379ec880126f4488)

Then look at the new config piece for annotation autodiscovery: [_annotation_autodiscovery.river.txt](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-fc581b5e7a75f916757d9a1a8f152744dd7541b98da2d8f641bb0e32a499d38c)

Look at the changes to support an annotation to skip logs auto gathering: [_pod_logs.river.txt](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-b987ac1d7de52caf887e3553ed845187d6123a3b98aacceef0126d6179a69ded)

Most of the other changes to the templates are for using a common function to turn k8s annotation names into prometheus-style labels (i.e. turn all the things into underscores).

The CI changes that automatically test integrations are:
* Config for Cert Manager chart: [cert-manager.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-b1733d03f414b4cef7766b25250c058eee28ad1f7164c76dc338f7659fd8fe4e)
* Config for MySQL chart: [mysql.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-f9d33bc6c759aa3aaae29207feb31e155a5495d46235d586404b2626920e962e)
* ConfigMap that holds the MySQL integration code (the snippet that we'd have users deploy): [mysql-config.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-5e99d8f6f7a4d4296f42863622ee1298de6f2fe2a14559d8934ad45dd8745ba3)
* [ci/ci-integrations-values.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/236/files#diff-6e37913fe5222027ed2965ec7f3533c3dbe37a622e0201e3f4349ef58066bb14)

The files in the examples directory will be mostly all the same, but might be worth seeing how the generated config is rendered.


charts/k8s-monitoring/README.md
```diff
@@ -141,6 +143,14 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely
...
 | metrics.apiserver.extraRelabelingRules | string | `""` | Rule blocks to be added to the discovery.relabel component for the API Server. See https://grafana.com/docs/agent/latest/flow/reference/components/discovery.relabel/#rule-block |
 | metrics.apiserver.scrapeInterval | string | 60s | How frequently to scrape metrics from the API Server Overrides metrics.scrapeInterval |
+| metrics.autoDiscover.annotations | object | `{"instance":"k8s.grafana.com/instance","job":"k8s.grafana.com/job","metricsPort":"k8s.grafana.com/metrics.port","scrape":"k8s.grafana.com/scrape"}` | Annotations that are used to discover and configure metric scraping targets. Add these annotations to your services or pods and the |
+| metrics.autoDiscover.annotations.instance | string | `"k8s.grafana.com/instance"` | Annotation for overriding the instance label |
+| metrics.autoDiscover.annotations.job | string | `"k8s.grafana.com/job"` | Annotation for overriding the job label |
+| metrics.autoDiscover.annotations.metricsPort | string | `"k8s.grafana.com/metrics.port"` | Annotation for setting or overriding the metrics port. Helpful if the port is not exposed on the Pod |
```
gaantunes, line 149: Should we have a metricsPath annotation as well? Sometimes it is not exposed on the default `/metrics` route.

petewall: Good call. I've added that in the most recent commit.

charts/k8s-monitoring/templates/agent_config/_annotation_autodiscovery.river.txt
```diff
@@ -0,0 +1,79 @@
...
+    action = "replace"
+    target_label = "__metrics_path__"
+  }
+  rule {
+    source_labels = [
+      "__address__",
+      "__meta_kubernetes_pod_annotation_{{ include "escape_label" .Values.metrics.autoDiscover.annotations.metricsPort }}",
+      "__meta_kubernetes_service_annotation_{{ include "escape_label" .Values.metrics.autoDiscover.annotations.metricsPort }}",
+    ]
+    regex = "(.+);;?([0-9]+);?"
```
skl, line 46: Is the double `;;` intentional?

skl (approved):
🥇

charts/k8s-monitoring/templates/agent_config/_annotation_autodiscovery.river.txt
petewall: Yeah, because we're trying to handle either the case of pods or services. If a pod matches, then the value will be `<address>;8080;` and if a service matches, it'll be `<address>;;8080`. The regex matches the address and the port, no matter if the port was set from the pod annotation or the service annotation.

petewall: FWIW, after chatting with Craig, I split that discovery.relabel component into two, one for services and one for pods. That'll clean things up a bunch.

skl (approved):
lgtm

rgeyer (approved):
Looks good! I like the level of configurability that's offered. Should cover most use-cases without too much drama.

Issue #261, opened 2023-11-23
When traces enabled, offer option to disable `disable_high_cardinality_metrics`


basvdl:
If you enable the trace receiver, the Agent will expose debug metrics like:

```
rpc_server_duration_milliseconds_bucket{component_id="otelcol.receiver.otlp.trace_receiver",net_sock_peer_addr="10.128.98.86",net_sock_peer_port="56388",otel_scope_name="go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc",otel_scope_version="0.45.0",rpc_grpc_status_code="0",rpc_method="Export",rpc_service="opentelemetry.proto.collector.trace.v1.TraceService",rpc_system="grpc",le="+Inf"} 1
rpc_server_duration_milliseconds_sum{component_id="otelcol.receiver.otlp.trace_receiver",net_sock_peer_addr="10.128.98.86",net_sock_peer_port="56388",otel_scope_name="go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc",otel_scope_version="0.45.0",rpc_grpc_status_code="0",rpc_method="Export",rpc_service="opentelemetry.proto.collector.trace.v1.TraceService",rpc_system="grpc"} 0
rpc_server_duration_milliseconds_count{component_id="otelcol.receiver.otlp.trace_receiver",net_sock_peer_addr="10.128.98.86",net_sock_peer_port="56388",otel_scope_name="go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc",otel_scope_version="0.45.0",rpc_grpc_status_code="0",rpc_method="Export",rpc_service="opentelemetry.proto.collector.trace.v1.TraceService",rpc_system="grpc"} 1
```

The longer the agent is running the more metrics are exposed, resulting in the agent consuming lots of memory and risking a OOM crash. The problem is the `peer_addr` and `peer_port` labels, causing high cardinality.

In my test the Agent was OOM killed every day when reaching 2.5GB memory. At one point I check the `/metrics` endpoint and counted 10764 metrics that started with `rpc_server_duration_milliseconds` and 1 hour later the same count resulted in 15516, and increase of 4752.

To disable this we should allow the user to set `debug_metrics` as part of the trace receiver config.

```
otelcol.receiver.otlp "trace_receiver" {
  grpc {
    endpoint = "[0.0.0.0:4317](http://0.0.0.0:4317/)"
  }
  output {
[...]

PR #262, opened 2023-11-23
Trace receiver disable debug metrics

basvdl:
Fixes https://github.com/grafana/k8s-monitoring-helm/issues/261

It also fixes the indentations for the `grpc` and `http` protocol blocks under the `otelcol.receiver.otlp`

### DIFF
--- a/charts/k8s-monitoring/Chart.yaml
+++ b/charts/k8s-monitoring/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: k8s-monitoring
 description: A Helm chart for gathering, scraping, and forwarding Kubernetes infrastructure metrics and logs to a Grafana Stack.
 type: application
-version: 0.5.2
+version: 0.5.1
 appVersion: 1.4.1
 icon: https://raw.githubusercontent.com/grafana/grafana/main/public/img/grafana_icon.svg
 sources:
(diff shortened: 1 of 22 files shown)


charts/k8s-monitoring/templates/agent_config/_traces.river.txt
```diff
@@ -6,16 +6,22 @@
 {{- if (or .Values.traces.receiver.grpc.enabled .Values.traces.receiver.http.enabled) }}
 otelcol.receiver.otlp "trace_receiver" {
 
+{{- if .Values.traces.receiver.disable_debug_metrics }}
```
skl, line 9: This looks correct for the oltp receiver but it also needs to be copied over to the zipkin receiver block as well.

basvdl: Gotcha.

Would we prefer to introduce 2 separate config items i.e.

```
.Values.traces.receiver. oltp .disable_debug_metrics
.Values.traces.receiver.zipkin.disable_debug_metrics
```

charts/k8s-monitoring/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: k8s-monitoring
 description: A Helm chart for gathering, scraping, and forwarding Kubernetes infrastructure metrics and logs to a Grafana Stack.
 type: application
-version: 0.5.1
+version: 0.5.2
```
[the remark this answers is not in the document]
basvdl: Cool, will revert these!

charts/k8s-monitoring/templates/agent_config/_traces.river.txt
skl: @basvdl that's probably the most robust option I guess, to have individual disable per receiver. Go for it.

skl (approved):
💯

PR #270, opened 2023-11-27
Remove the filename label from logs

### LINKED ISSUE #267: Unable to drop the "filename" label
I've tried  dropping the filename label which has a very high cardinality in my project (running dozens of ephemeral pods per minute)

I have tried multiple different ways but I'm unable to drop the `filename` and `stream` label.

Here is an excerpt from my helm values:
```yml
logs:
  enabled: true
  pod_logs:
    namespaces:
      - company_name
      - default
    extraRelabelingRules: |
      rule {
        regex = "^(pod|job|filename|container|stream)$"
        action = "labeldrop"
      }
      rule {
        regex = "^filename$"
        action = "labeldrop"
      }
      rule {
        source_labels = ["filename"]
        regex = "/.+"
        action = "drop"
      }
``` 

How do I get rid of `filename`

petewall:
Fixes #267

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_pod_logs.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_pod_logs.river.txt
@@ -100,9 +100,11 @@ loki.process "pod_logs" {
     }
   }
 
-  // drop the temporary container runtime label as it is no longer needed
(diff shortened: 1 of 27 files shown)


skl (approved):
I tested on a stack and this works great, thanks! 💯

PR #284, opened 2023-12-07
Introduce a configuration analyzer into Helm test which shows information about how discovery and scraping are performing

petewall:
Also, build a test image, rather than installing the dependencies over and over.


petewall:
Haha. Now the CI is failing, not because the code is bad, but because the `apt-get update && apt-get install -y curl jq` slowed the tests long enough for the metrics to show up.

charts/k8s-monitoring/test/config-analysis.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+set -eo pipefail
+
+AGENT_HOST="${AGENT_HOST:-http://localhost:8080}"
+sources=$1
+
+function discoveryRelabel() {
+    component=$1
```
skl, line 9: May not be necessary but I think there might be some scope leakage here, so you could use `local`?

```suggestion
    local component=$1
```

charts/k8s-monitoring/test/config-analysis.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+set -eo pipefail
+
+AGENT_HOST="${AGENT_HOST:-http://localhost:8080}"
+sources=$1
```
skl, line 6: Maybe a `test -z` or similar if this is a required arg?

skl (approved):
💯

charts/k8s-monitoring/test/config-analysis.sh
petewall: Good call.

PR #295, opened 2023-12-18
Make helm test work with the --logs flag

petewall:
Need to convert the jobs to pods and stop them from being auto-deleted.

### DIFF
--- a/charts/k8s-monitoring/templates/tests/test.yaml
+++ b/charts/k8s-monitoring/templates/tests/test.yaml
@@ -89,7 +89,7 @@ spec:
     {{- toYaml .Values.test.image.pullSecrets | nindent 8 }}
     {{- end }}
   {{- end }}
-  restartPolicy: Never
+  restartPolicy: OnFailure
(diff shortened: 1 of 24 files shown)


charts/k8s-monitoring/templates/tests/test.yaml
```diff
@@ -75,57 +75,39 @@ metadata:
     {{- end}}
   annotations:
     "helm.sh/hook": test
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-delete-policy": before-hook-creation
```
skl, line 78: Is any cleanup required by the user after running helm test?

skl (approved):

charts/k8s-monitoring/templates/tests/test.yaml
petewall: They just... linger there. I don't love it.
I'm not sure I'm happy with this PR as it stands.

PR #303, opened 2023-12-19
When determining when to run the tests, check against the latest release, not against main

petewall:
`git describe --abbrev=0 --tags` gets the latest release (as a git tag).
`ct list-changes` takes a `--since` which uses a git tag.

### DIFF
--- a/.github/workflows/helm-test.yml
+++ b/.github/workflows/helm-test.yml
@@ -124,7 +124,8 @@ jobs:
     - name: List changed charts
       id: list-changed
       run: |
-        changed=$(ct list-changed --config "${CT_CONFIGFILE}")
+        latestRelease=$(git describe --abbrev=0 --tags)
+        changed=$(ct list-changed --config "${CT_CONFIGFILE}" --since "${latestRelease}")
         if [[ -n "$changed" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-test.yml
```diff
@@ -124,7 +124,8 @@ jobs:
     - name: List changed charts
       id: list-changed
       run: |
-        changed=$(ct list-changed --config "${CT_CONFIGFILE}")
+        latestRelease=$(git describe --abbrev=0 --tags)
+        changed=$(ct list-changed --config "${CT_CONFIGFILE}" --since "${latestRelease}")
```
skl, line 128: This same workflow runs on both `main` and PRs, right? So does this mean that this case is always true for every commit made on every branch since the last release... i.e. it can never be false except for the release commit itself?

skl (approved):
Approving anyway because A) I trust your judgement and B) always running on main is better than never! 😄

.github/workflows/helm-test.yml
petewall: It actually checks if the chart itself has changed. So a change to an example or a change to the top-level README.md wont trigger a test.

PR #319, opened 2024-01-03
Extract Kubernetes Event gathering into its own Agent deployment

petewall:


petewall:
After:
image

```text
% helm test k8smon --logs
...
POD LOGS: test-k8smon-k8s-monitoring
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=up{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=agent_build_info{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=kubernetes_build_info{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=machine_memory_bytes{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=kube_node_info{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=node_exporter_build_info{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=apiserver_request_total{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=workqueue_adds_total{cluster="local-test", job="kube-controller-manager"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=kubeproxy_sync_proxy_rules_service_changes_total{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=scheduler_unschedulable_pods{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=opencost_build_info{cluster="local-test"}...
Running PromQL query: https://prometheus-server.prometheus.svc:80/api/v1/query?query=grafana_kubernetes_monitoring_build_info{cluster="local-test"}...
Running LogQL query: http://loki.loki.svc:3100/loki/api/v1/query?query={cluster="local-test", job!="integrations/kubernetes/eventhandler"}...
Running LogQL query: http://loki.loki.svc:3100/loki/api/v1/query?query={cluster="local-test", job="integrations/kubernetes/eventhandler"}...
All queries passed!
```

.github/configs/grafana.yaml
```diff
@@ -4,7 +4,7 @@ datasources:
     datasources:
       - name: Prometheus
         type: prometheus
-        url: https://prometheus-server.prometheus.svc:80/prometheus
+        url: https://prometheus-server.prometheus.svc:80
```
petewall, line 7: This change is just for fixing the local cluster test environment

scripts/setup-local-test-cluster.sh
```diff
@@ -9,7 +9,7 @@ GRAFANA_VALUES="./.github/configs/grafana.yaml"
 SECRETGEN_CONTROLLER_MANIFEST=https://github.com/carvel-dev/secretgen-controller/releases/latest/download/release.yml
 CERTIFICATES_MANIFEST="./.github/configs/certificates.yaml"
 
-if ! kind get nodes --name "${CLUSTER_NAME}" > /dev/null 2>&1; then
+if ! kind get nodes --name "${CLUSTER_NAME}" | grep "No kind nodes found for cluster \"${CLUSTER_NAME}\"" > /dev/null 2>&1; then
```
petewall, line 12: This also is for fixing the local cluster test environment. Turns out `kind get nodes` does not "error" when the cluster does not exist.

charts/k8s-monitoring/Chart.yaml
```diff
@@ -16,6 +16,11 @@ dependencies:
 - name: grafana-agent
   version: 0.29.0
   repository: https://grafana.github.io/helm-charts
+- alias: grafana-agent-events
```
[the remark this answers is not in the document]
petewall: Done

charts/k8s-monitoring/templates/_configs.tpl
```diff
@@ -0,0 +1,105 @@
...
+    {{- include "agent.config.traces" . }}
+    {{- include "agent.config.tracesService" . }}
+  {{- end }}
+
+  {{- if .Values.extraConfig }}
+    {{- print "\n" .Values.extraConfig }}
+  {{- end }}
+{{- end -}}
+
+{{/* Grafana Agent Events config */}}
```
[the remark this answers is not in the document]
petewall: I don't think so. At least not pro-actively.
There are lots of methods for adding extra config. I'll be happy to wait until someone asks for it.

examples/default-values/output.yaml
```diff
@@ -1,4 +1,17 @@
 ---
+# Source: k8s-monitoring/charts/grafana-agent-events/templates/serviceaccount.yaml
+apiVersion: v1
+kind: ServiceAccount
```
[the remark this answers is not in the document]
petewall: Yes! Thank you. Updated.

charts/k8s-monitoring/templates/grafana-agent-events-config.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if and .Values.logs.enabled .Values.logs.cluster_events.enabled }}
```
sidewinder12s, line 1: This might have a bug, it appears if you have logs disabled but cluster_events enabled, the configmap doesn't get generated and then the agent-events deployment fails to start. At least in our case we did not want the daemonset logging deployment but we had hoped to collect events with Grafana.

petewall: Yeah.. it's an inherited mess that I hope to untangle in the future.

the reason cluster_events is inside of logs is because they're processed as log lines and stored in loki. If I could do it again, they'd be at the top level. So `logs.cluster_events.enabled` is the switch for events, `logs.pod_logs.enabled` is the switch for pod logs, and `logs.enabled` is the switch for both.

If you change your config to use:
```yaml
logs:
  pod_logs:
    enabled: false
  cluster_events:
    enabled: true
```

then things should work for you.

petewall: Eventually, I may remove `logs.enabled` or even move `cluster_events` outside of `logs`, but that'll come with deprecation warnings before I do.

reberhardt7:
Hi! Thanks so much for all your work on this project. It has made it much easier to collect telemetry to Grafana in our k8s cluster.

Just trying to understand context here: What's the advantage of splitting these into separate Agent deployments? Why not have one agent deployment that collects metrics, logs, and cluster events?

petewall:
Great question! It's all about ensuring the telemetry data is gathered efficiently, but also without duplications.
Metrics are gathered with a stateful set, and the Grafana Agent supports clustering. If the agent pods are using too much resources, they can be horizontally scaled to distribute the load
Logs are gathered with a daemonset, since it reads the files directly from the node.
Events are gathered with a connection to the k8s api.

If we put all of the logic into a single agent deployment, and assume we scaled to multiple pods, we'd risk either missing data (because the pods wouldn't be on all node) or gathering duplicates (event gathering can only be on one node at the moment).

Splitting the work across three separate Grafana Agent deployments means that each can be deployed, configured, and scaled, optimally to handle the work that they have to do.

PR #321, opened 2024-01-03
Allow for metric relabeling rules for operator objects

petewall:
Also update the schema


eamonryan (approved):
:+1:

charts/k8s-monitoring/values.yaml
```diff
@@ -642,6 +642,10 @@ metrics:
     # @default -- 60s
     scrapeInterval: ""
 
+    # -- Rule blocks to be added to the prometheus.relabel component for the Kube Controller Manager.
```
sidewinder12s, line 645: Update the comment for the following 3

petewall: Ooh! Thanks for catching that.

PR #323, opened 2024-01-05
Add action to build the test image

petewall:


.github/workflows/update-test-image.yml
```diff
@@ -0,0 +1,39 @@
...
+        id: get-version
+        run: yq -r '"version=\(.version)"' source/charts/k8s-monitoring/Chart.yaml >> $GITHUB_OUTPUT
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/grafana/k8s-monitoring-test
+          tags: |
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'master') }}
```
petewall, line 25: Yes, this will always be true, but I want to keep this as a breadcrumb if we expand this to build images tagged for PRs.

.github/workflows/update-test-image.yml
```diff
@@ -0,0 +1,39 @@
...
+        id: get-version
+        run: yq -r '"version=\(.version)"' source/charts/k8s-monitoring/Chart.yaml >> $GITHUB_OUTPUT
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/grafana/k8s-monitoring-test
+          tags: |
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
```
petewall, line 25: Yes, this will always be true, but I'd rather leave this as a breadcrumb if we ever expand this to build images based on PRs.

skl (approved):
Nice

Issue #46, opened 2023-08-01
Add sane defaults to reduce cardinality


davhdavh:
The current config will generate close to 10.000 metrics on an near empty cluster.

Much of the problem is that something like the cadvisor scrape for networking will generate a series per pod per interface per device per xxx easily generating 1000-2000 unique metrics.

The reality is that by far most of these unique labels are not used.

davhdavh:
What I am thinking here, is a list of predefined sets of metrics reduction:
* All (current)
* Pod (should be default, expecting ~50% reduction in unique metrics series)
* Workload (pod level metrics removed, 75% reduction)
* Namespace (workload level metrics remove, 90% reduction)

some of the charts are using some secret metrics. e.g. `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate` is there anywhere I can find the specification on this?

davhdavh:
Preliminary testing on a single node cluster shows a bit better than 50% savings without loss of any charts.
WITH POD-level rules: 6.94k unique metrics, 5.80k as k8s intg, 122 as linux intg.
WITHOUT rules: 13.7k unique metrics, 11.9k as k8s intg, 796 as linux intg. - but logs say it gets rate limited at 15k (im just testing on a free account)
I'm fairly certain we can cut another 10-20% without loss of charts by focusing on the allowLists a bit more.

davhdavh:
Rules used
[custom_rules.zip](https://github.com/grafana/k8s-monitoring-helm/files/12249711/custom_rules.zip)

davhdavh:
Workload level rules (I changed a bit, so pod-level is slightly different from above)
WITH POD-level rules: 6.23k unique metrics, 5.11k as k8s intg, 105 as linux intg.
WITH workload-level rules: 4.55k unique metrics, 3.52k as k8s intg, 105 as linux intg.
This WILL break some charts, but if you just care about an overview of the metrics at low cost, that might be what you want.

davhdavh:
WITH namespace-level rules: 1.62k unique metrics, 1.0k as k8s intg, 105 as linux intg.

petewall:
This is very interesting! Thanks for digging into this.

The "secret metrics" are recording rules are set from this repo: https://github.com/grafana/kubernetes-mixin. If you search there, you can find the metrics that are used by that rule.

davhdavh:
Yeah. I found them in the UI. Recording rules they are called and for some odd reason specified under alerts.

[...]

PR #326, opened 2024-01-05
Add more options for metrics tuning

petewall:
Implementing some of the incredible work done by @davhdavh in this thread: https://github.com/grafana/k8s-monitoring-helm/issues/46

We may not take all immediately, but I'm creating a PR to start testing the ideas.


davhdavh:
Nice. Looking forward to it

petewall:
@skl , would you review the latest, especially the changes to KSM?
I'm trying out putting our default allow lists into the cardinality improvements, with an includeMetrics and excludeMetrics options that will add or remove metrics from the allow list.

A few options that would work:
1. Default, no changes to values: No change from today, default allowList
2. allowList disabled:
```
metrics:
  kube-state-metrics:
    cardinalityImprovements:
      useAllowList: false
```
All metrics will be sent

3. Allow list with include and exclude
```
metrics:
  kube-state-metrics:
    cardinalityImprovements:
      useAllowList: true
      includeMetrics: ["kube_pod_labels"]
      excludeMetrics: ["kube_pod_info"]
```
It'll use the allowList, but also allow `kube_pod_labels`, but `kube_pod_info` will be dropped

4. No allow list, but include list
```
metrics:
  kube-state-metrics:
    cardinalityImprovements:
      useAllowList: false
      includeMetrics: ["kube_node_info"]
```

whole new allow list. ONLY `kube_node_info` is sent

petewall:
One reason I like embedding the allow list as a switch, it also means that we can include options for alternative allow lists.
For example, in Node Exporter, we can default to our current allowList, but include a second allow list that will make it work with the Node Exporter integration.

Same for the Grafana Agent integration. Currently, we just return "agent_build_info", but we could set the switch to include the metrics that the agent integration dashboards need.

charts/k8s-monitoring/README.md
```diff
@@ -242,6 +250,7 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely
 | metrics.kubelet.scrapeInterval | string | 60s | How frequently to scrape metrics from the Kubelet. Overrides metrics.scrapeInterval |
 | metrics.kubernetesMonitoring.enabled | bool | `true` | Report telemetry about this Kubernetes Monitoring chart as a metric. |
 | metrics.node-exporter.allowList | list | See [Allow List for Node Exporter](#allow-list-for-node-exporter) | The list of Node Exporter metrics that will be scraped by the Agent |
+| metrics.node-exporter.cardinalityImprovements.dropTempFSMetrics | bool | `true` | Drop node filesystem metrics with tempfs |
```
skl, line 253: I like the more configurable version in `metrics.cadvisor.cardinalityImprovements.keepPhysicalFilesystemDevices`, rather than this simpler option.

charts/k8s-monitoring/values.yaml
```diff
@@ -292,37 +292,22 @@ metrics:
...
-    - kube_replicaset.*
-    - kube_resourcequota
-    - kube_statefulset.*
+    cardinalityImprovements:
+      # -- Filter the list of metrics from Kube State Metrics to a useful, minimal set. See [Allow List for Kube State Metrics](#allow-list-for-kube-state-metrics)
+      useAllowList: true
+      # -- Metrics to be kept
+      includeMetrics: []
+      # -- Metrics to be dropped
+      excludeMetrics: []
```
skl, line 301: What's the default behaviour if these are set to `[]`? Might be worth a comment to clarify.

skl (commented):
Nice, I like the direction this is going!

charts/k8s-monitoring/README.md
petewall: Done

charts/k8s-monitoring/values.yaml
petewall: Good call. I'll add some better comments in a future commit.

charts/k8s-monitoring/values.yaml
```diff
@@ -228,6 +228,12 @@ metrics:
     # See https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.relabel/#rule-block
     extraMetricRelabelingRules: ""
 
+    cardinalityImprovements:
```
skl, line 231: Just a thought but `cardinalityImprovements` also can increase cardinality depending on the values, e.g. by including more metrics. Suggest maybe `cardinalityModifications` or simply `cardinality`?

charts/k8s-monitoring/templates/agent_config/_agent.river.txt
```diff
@@ -52,13 +61,20 @@ prometheus.scrape "agent" {
...
     action = "keep"
   }
 {{- end }}
+{{- if .Values.metrics.agent.metricsTuning.excludeMetrics }}
+  rule {
+    source_labels = ["__name__"]
+    regex = {{ join "|" .Values.metrics.agent.metricsTuning.excludeMetrics | quote }}
+    action = "drop"
+  }
+{{- end }}
```
jewbetcha, line 77: Nice, this will be helpful

jewbetcha (approved):

petewall:
Putting back to draft mode. Things I still need to change:
* Add a deprecation warning in the README. This will affect anyone who is using a custom allow list
* Add a check to fail deployments if their values file includes an allowList.

charts/k8s-monitoring/values.yaml
```diff
@@ -303,37 +316,23 @@ metrics:
...
+      excludeMetrics: []
+      # -- Normalize labels to the same value for the given metric and label pairs
+      normalizeUnnecessaryLabels:
+        - labels: ["image", "image_id", "image_spec"]
+        - metric: "kube_pod_info|kube_pod_owner"
+          labels: ["created_by_name"]
+        - metric: "kube_pod_container_info"
+          labels: ["container_id"]
+        - metric: "kube_pod_info"
+          labels: ["host_ip", "pod_ip"]
```
petewall, line 335: I need to make sure that renaming these labels to be the same will not break anything on the K8s Mon plugin
(cc @skl @jewbetcha)

charts/k8s-monitoring/values.yaml
```diff
@@ -475,27 +450,28 @@ metrics:
...
+      # -- Normalize labels to the same value for the given metric and label pairs
+      normalizeUnnecessaryLabels:
+        - metric: "container_cpu_.*|container_fs_.*|container_memory_.*|container_network_.*"
+          labels: ["id", "name", "image"]
+        - metric: "machine_memory_bytes"
+          labels: ["boot_id", "system_uuid"]
+      # -- Only keep filesystem metrics that use the following physical devices
+      keepPhysicalFilesystemDevices: ["mmcblk.p.+", "nvme.+", "rbd.+", "sd.+", "vd.+", "xvd.+", "dasd.+"]
+      # -- Only keep network metrics that use the following physical devices
+      keepPhysicalNetworkDevices: ["en[ospx][0-9].*", "wlan[0-9].*", "eth[0-9].*"]
```
petewall, line 474: Same with these

joejulian:
Ah, very nice. This is what I've been prowling through issues looking how to do! :heart:

khaykingleb:
Really waiting for this stuff! Thanks for your work :pray:

charts/k8s-monitoring/templates/agent_config/_node_exporter.river.txt
```diff
@@ -50,13 +176,29 @@ prometheus.scrape "node_exporter" {
 }
 
 prometheus.relabel "node_exporter" {
-{{- if (index .Values.metrics "node-exporter").allowList }}
+{{- if or (index .Values.metrics "node-exporter").metricsTuning.useDefaultAllowList (index .Values.metrics "kube-state-metrics").metricsTuning.includeMetrics }}
```
skl, line 179: Did you mean to use KSM here?

skl (approved):
🚀

PR #274, opened 2023-11-28
Support using receivers for Metrics, Logs, and Traces

petewall:
Definition: A receiver: The combination of an open port and configuration that allows the Grafana Agent to be pushed data, rather than scraping it (pulling data).

Currently, traces use receivers, but metrics and logs do not. In order to further support more flexible deployment patterns, this PR will generalize receivers so they can receive all MLT and send them to their appropriate back-ends.

Notes to reviewers:
The receivers that used to be inside of `traces` (i.e. `traces.receivers`) are now at the top level (i.e. `receivers`)

As always, start with the `charts/k8s-monitoring/templates` directory.

This also includes code that deploys a Grafana Agent in CI that's only there to generate metrics and traces and send them to the K8s Monitoring chart, using the OTLP and remote_write receivers.

PR #342, opened 2024-01-16
add k8s.cluster.name resource attribute

mar4uk:
This PR is a successor of https://github.com/grafana/k8s-monitoring-helm/pull/219 taking into account changes made in https://github.com/grafana/k8s-monitoring-helm/pull/274

@cedricziel explained very well why we need this change in his initial PR: https://github.com/grafana/k8s-monitoring-helm/pull/219#issuecomment-1787420167

TLDR;
We need spans to have resource attribute `k8s.cluster.name` to enable navigation between app o11y and k8s monitoring

### DIFF
--- a/.github/configs/agent-config.yaml
+++ b/.github/configs/agent-config.yaml
@@ -31,20 +31,26 @@ agent:
       
       tracing {
         sampling_fraction = 0.1
-        write_to = [otelcol.processor.attributes.trace_attributes_processor.input]
+        write_to = [otelcol.processor.transform.trace_transform_processor.input]
       }
       
-      otelcol.processor.attributes "trace_attributes_processor" {
-        action {
-          key = "cluster"
-          value = "from-the-other-agent"
-          action = "insert"
+      otelcol.processor.transform "trace_transform_processor" {
+        error_mode = "ignore"
+
+        trace_statements {
+          context = "resource"
+          statements = [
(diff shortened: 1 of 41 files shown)


petewall (commented):
In general, I like these changes. They make sense given the discussion we had on Slack.
In specific, I'd prefer that the assignment of the `k8s.cluster.name` be still done in _traces.river.txt.

We can rename `otelcol.processor.attributes "trace_attributes_processor"` to `otelcol.processor.transform "trace_transform_processor"`.

the Receivers file should just contain code for the receivers, while the specific code that modifies traces should stay in the traces file.

This way, we also get the `cluster` variable, too.

cyrille-leclerc:
@petewall my understanding was that you shipped before Christmas a change to make the OTLP endpoint not only support traces but also metrics and logs. If it's the case then would `k8s.cluster.name` not only be added to OTLP traces but also to OTLP metrics and logs?

petewall:
Yeah, we absolutely should add that to all OTLP inputs.

petewall (approved):

mar4uk:
Hey @petewall thank you for your review! I made changes to address your feedback
I would add `k8s.cluster.name` resource attribute to other signals in another PR (I'm still figuring out if it is needed for app o11y to work)

mar4uk:
@petewall could you please give me some hints as to why tests are failing? Is it related to my changes or it is something else, I have difficulty finding out what is the reason for the failure

PR #351, opened 2024-01-25
Improve how the log gathering annotation behaves.

petewall:
We can configure the Grafana Agent to gather logs in two ways (by setting `logs.pod_logs.discovery`

## `logs.pod_logs.discovery=all`

Currently, logs from all pods will be gathered, except when the annotation `k8s.grafana.com/logs.autogather: skip`
After the change, the value of the annotation can be `false`, `no`, or `skip`

## `logs.pod_logs.discovery=annotation`

Currently, only logs from pods with the annotation `k8s.grafana.com/logs.autogather: gather` will be gathered.
After the change, the value of the annotation can be anything *except* `false`, `no`, or `skip`


jewbetcha (approved):
Thanks for the detailed explanation in the PR 🙏

skl (approved):

PR #353, opened 2024-01-26
Extract the allow lists into their own files so they're easier to find and inspect

petewall:
The default allow lists were embedded in Helm chart templates. If we extract them into flat YAML files, they're easier to reason about. We can then dynamically include them at Helm template time.

Eventually, this could be used to automatically update the README.md, but that's a level of `sed` skills that I don't yet have.


petewall:
Adding you, @BeverlyJaneJ , because we'll be able to link to these files.

skl (approved):

PR #354, opened 2024-01-26
Make it so helm test queries can assert values

petewall:
This change adds the ability to support an `expect` section in the test queries.
For example:
```
  - query: "count(kube_pod_info{cluster=\"ci-test-cluster-2\", color=\"blue\"})"
    type: promql
    expect:
      operator: ">"
      value: 1
```
Will not only check that the query returns successfully, but will assert that the value is greater than 1.

Also supported are counting the number of results, such as:

```
metrics:
  kube-state-metrics:
    metricsTuning:
      excludeMetrics:
        - kube_pod_info

test:
  attempts: 20
  extraQueries:
    # Check that the metrics exclusion list is working
    - query: "kube_pod_info{cluster=\"ci-test-cluster\"}"
      type: promql
      expect:
        count: 0
```
Which allows us to do negative testing for the first time.

### DIFF
--- a/charts/k8s-monitoring/ci/ci-values.yaml
+++ b/charts/k8s-monitoring/ci/ci-values.yaml
@@ -41,6 +41,10 @@ metrics:
     enabled: true
   kubeScheduler:
     enabled: true
+  kube-state-metrics:
+    metricsTuning:
+      excludeMetrics:
+        - kube_pod_info
 
 traces:
   enabled: true
@@ -52,8 +56,12 @@ test:
     - query: "count(kube_node_info{cluster=\"ci-test-cluster\", color=\"red\"})"
       type: promql
       expect:
-        operator: ==
         value: 1
+    # Check that the metrics exclusion list is working
+    - query: "kube_pod_info{cluster=\"ci-test-cluster\"}"
+      type: promql
+      expect:
+        count: 0
     # Check for cluster events
     - query: "{cluster=\"ci-test-cluster\", job=\"integrations/kubernetes/eventhandler\"}"
       type: logql
--- a/charts/k8s-monitoring/ci/ci-values.yaml
+++ b/charts/k8s-monitoring/ci/ci-values.yaml
@@ -49,8 +49,11 @@ test:
   attempts: 20
   extraQueries:
     # Check that the external label is properly being set
-    - query: "kube_node_info{cluster=\"ci-test-cluster\", color=\"red\"}"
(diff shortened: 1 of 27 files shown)


charts/k8s-monitoring/ci/ci-values.yaml
```diff
@@ -49,8 +53,15 @@ test:
   attempts: 20
   extraQueries:
     # Check that the external label is properly being set
-    - query: "kube_node_info{cluster=\"ci-test-cluster\", color=\"red\"}"
+    - query: "count(kube_node_info{cluster=\"ci-test-cluster\", color=\"red\"})"
       type: promql
+      expect:
+        value: 1
```
jewbetcha, line 59: This only runs on the test cluster right? So we always know there will be 1 node?

charts/k8s-monitoring/test/query-test.sh
```diff
@@ -51,10 +137,13 @@ count=$(jq -r ".queries | length-1" "${1}")
 for i in $(seq 0 "${count}"); do
   query=$(jq -r --argjson i "${i}" '.queries[$i].query' "${1}")
   type=$(jq -r --argjson i "${i}" '.queries[$i] | .type // "promql"' "${1}")
+  expectedCount=$(jq -r --argjson i "${i}" '.queries[$i].expect.count // empty | tostring' "${1}")
+  expectedValue=$(jq -r --argjson i "${i}" '.queries[$i].expect.value // empty | tostring' "${1}")
+  expectedOperator=$(jq -r --argjson i "${i}" '.queries[$i].expect | .operator // "=="' "${1}")
```
jewbetcha, line 142: I wonder if we should/could do the query testing in a different language (like Go) for longevity sake. Something to think about.

jewbetcha (approved):
LGTM. Just some questions in there

charts/k8s-monitoring/ci/ci-values.yaml
petewall: Yeah, this is the values run during CI.

charts/k8s-monitoring/test/query-test.sh
petewall: Perhaps.
A reason I like Bash for this is that I have run this script directly before. Just export some environment vars and go.
It makes things very portable without having to compile it.

jewbetcha: Sure, that makes sense

skl (approved):

PR #360, opened 2024-02-01
fix: use the port of the annotation when specified

daquinoaldo:
## Summary

This change aims at fixing an issue where the pod is not scraped despite the `k8s.grafana.com/metrics.portNumber` annotation (or equivalent annotation specified in the `metrics.autoDiscover.annotations.metricsPortNumber` value, such as `promethus.io/port`), when the Pod doesn't declare that port among the container ports.

### Why this is important

**All customers using Istio are affected.**
When the Istio controller mutates the pod to inject the istio-proxy container, it reads the `prometheus.io/port` value. It uses it to scrape application metrics from the main container and re-expose the [merged metrics](https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-metrics-merging) (istio-proxy Envoy metrics + main container metrics) on port 15020, which is set as the value for the `prometheus.io/port` annotation. However, the `istio-proxy` container only declares port 15090, exposing Envoy-only metrics.
This behavior is well explained in an [issue](https://github.com/istio/istio/issues/47882) on the Istio repository.

#### What happens with the Grafana Agent

As in Prometheus, the pod discovery of the Grafana Agent generates a target for each declared container port of the pod.
The Grafana Agent configurations of this Chart then discard all the pod-port couples whose port doesn't match the one in the annotations. Theoretically, only a couple should survive, the one with the port specified in the annotation. However, if that port is not declared, that target doesn't exist in all possible couples, and none survives. Consequently, all the pods part of the Istio mesh (all but a few system pods) are dropped from the scrape targets.

## Proposed change

When the `metricsPortNumber` annotation value is set, always use that port for the scraping, ignoring the declared container ports of the pod.

This is the [current Prometheus approach](https://github.com/prometheus-community/helm-charts/blob/cd640243eb03787fab11cf2883534f604787f760/charts/prometheus/values.yaml#L1068-L1077) and ensures the pod is scraped even if the container port is not declared.

### Detailed explanation

- If the annotation specified in the `metricsPortName` value exists on the pod, then discard all couples but that port.
  - I didn't change this behavior because the annotation specifies the port name, so the only way to get the number is by looking at declared ports.
  - If the pod doesn't specify that label, no action is taken: the [regex](https://github.com/grafana/k8s-monitoring-helm/blob/469eceee5e61c5a2fea9c4a914974d83cf6c1252/charts/k8s-monitoring/templates/agent_config/_annotation_autodiscovery.river.txt#L33) doesn't match and as per [doc](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config:~:text=If%20regex%20does%20not%20match%2C%20no%20replacement%20takes%20place.) the replace doesn't happen. Consequently, `__tmp_port` retains the previous value, equal to `__meta_kubernetes_pod_container_port_name`, and the `keepequal` doesn't discard the couple.
  - Istio doesn't require such annotation, so customers using Istio only need not to specify it and won't be affected by this rule.
- If the annotation specified in the `metricsPortName` value exists on the pod, then replace `__address__` with `podIp:port`, where the port is the annotation value.
  - In this way, all the pod-port couples will be replaced with the same value, and since they are duplicates, only one is kept. Thus, the final result is the same: only the target with the port of the annotation survives.
  - When the port from the annotation is not in the declared port, now the target is correctly set.
  - _If a container has no specified ports, a port-free target per container is created_ ([ref](https://grafana.com/docs/agent/latest/flow/reference/components/discovery.kubernetes/#pod-role)). Previously, _These targets must have a port manually injected using a discovery.relabel[ component](https://grafana.com/docs/agent/latest/flow/reference/components/discovery.relabel/) before metrics can be collected from them._ Now, if the port annotation is specified, the correct target is already generated without additional rules.

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_annotation_autodiscovery.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_annotation_autodiscovery.river.txt
@@ -24,6 +24,8 @@ discovery.relabel "annotation_autodiscovery_pods" {
   }
 
   // Choose the pod port
+  // The discovery generates a target for each declared container port of the pod.
+  // If the metricsPortName annotation has value, keep only the target where the port name matches the one of the annotation.
   rule {
     source_labels = ["__meta_kubernetes_pod_container_port_name"]
     target_label = "__tmp_port"
@@ -39,19 +41,19 @@ discovery.relabel "annotation_autodiscovery_pods" {
     target_label = "__tmp_port"
   }
 
+  // If the metricsPortNumber annotation has value, override the target address to use it,
+  // independently from the declared container ports of the pod.
   rule {
-    source_labels = ["__meta_kubernetes_pod_container_port_number"]
-    target_label = "__tmp_port"
+    source_labels = ["__meta_kubernetes_pod_annotation_{{ include "escape_label" .Values.metrics.autoDiscover.annotations.metricsPortNumber }}", "__meta_kubernetes_pod_ip"]
+    regex = "(\\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})"
+    replacement = "[$2]:$1" // IPv6
+    target_label = "__address__"
   }
   rule {
-    source_labels = ["__meta_kubernetes_pod_annotation_{{ include "escape_label" .Values.metrics.autoDiscover.annotations.metricsPortNumber }}"]
-    regex = "(.+)"
-    target_label = "__tmp_port"
-  }
-  rule {
-    source_labels = ["__meta_kubernetes_pod_container_port_number"]
-    action = "keepequal"
-    target_label = "__tmp_port"
+    source_labels = ["__meta_kubernetes_pod_annotation_{{ include "escape_label" .Values.metrics.autoDiscover.annotations.metricsPortNumber }}", "__meta_kubernetes_pod_ip"]
+    regex = "(\\d+);((([0-9]+?)(\\.|$)){4})" // IPv4, takes priority over IPv6 when both exists
+    replacement = "$2:$1"
+    target_label = "__address__"
   }
 
   rule {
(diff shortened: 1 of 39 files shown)


petewall:
I like the idea of this. In fact, during initial development of using annotations to discover and scrape targets, I wanted to be able to override targets. The trouble comes in with preventing multiple scrapes of the same target.

The `pod` discovery will return a target for every port in every container in every pod.
That means if there is a pod with two containers, and each container exposes two ports, then you'll have four targets to choose from.

Normally, this isn't so bad, because if only one of those ports returns prometheus metrics, we'll scrape those and then ignore the other three ports that don't return what we expect.

However if we force the port to be the metrics port, then we'll have four targets that all point to the same location. They'll all get scraped, processed, and shipped to Prometheus (or Mimir, or wherever). You may never know that you're actually generating 4x the metrics load that you expect.

petewall:
Ok, good news! After talking with the Agent team, it looks like as long as all public labels are the same (i.e. nothing prefixed with `__`), the Agent will dedupe the target list before scraping the metrics.

That means that this approach will likely work just fine. Let me review and give guidance for how to proceed.

petewall (approved):

petewall:
The changes look good. In order to pass the tests, please run:
```
make regenerate-example-outputs
```

in the root level of the repository. That will update the examples to include your changes.

daquinoaldo:
Hi @petewall, thanks for your prompt review.

Yes, I've noticed the Grafana Agent uses the Prometheus service discovery, which deduplicates targets.

I regenerated example outputs, you can re-run tests.

daquinoaldo:
Thank you! Do you have an ETA for the release?

petewall:
Prepping 0.9.3 right now!

petewall:
When this is complete, I'll push the release button: https://github.com/grafana/k8s-monitoring-helm/actions/runs/7758024065

petewall:
Released! Thank you so much for your contribution.

PR #370, opened 2024-02-09
Expand default allowlist agent

duncan485:
Should the default allowlist be expanded so it follows the config for the grafana agent health integration ? 

https://grafana.com/docs/grafana-cloud/monitor-infrastructure/integrations/integration-reference/integration-grafana-agent/#metrics


petewall:
Thanks for adding this PR! I'm concerned with how many additional metrics this will generate for users that don't want agent info, so what I think we should do is mirror what we do for Node Exporter.

If you look inside `charts/k8s-monitoring/default_allow_lists`, you'll see that we have two allowlists for Node Exporter, one is the default that we need for Kubernetes Monitoring, the other is what's required for the Linux node integration.

I propose that you move the metric list into a new file, `agent_integration.yaml`, then adjust `charts/k8s-monitoring/templates/agent_config/_agent.river.txt` to be able to pick up either.

Take a look at https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/templates/agent_config/_node_exporter.river.txt#L2-L7 for how we pick between the files, and https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/values.yaml#L359-L363 to see how we define it in the values.yaml file.

Finally, don't forget to regenerate the docs and examples (https://github.com/grafana/k8s-monitoring-helm/blob/main/README.md#contributing)

duncan485:
Hi @petewall, Thanks for the quick response, sounds good to me. I'll have a look once I have time !

duncan485:
Hi @petewall the values file actually already had a .metrics.agent.metricsTuning.useIntegrationAllowList https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/values.yaml#L285. But it just wasn't used for the river config yet. I updated template and added the seperate file. But no updates to docs or values.yaml, because it was already there.

charts/k8s-monitoring/templates/agent_config/_agent.river.txt
```diff
@@ -1,7 +1,11 @@
 {{ define "agent.config.agent.allowList" }}
-{{ if .Values.metrics.agent.metricsTuning.useDefaultAllowList }}
+{{ if or (index .Values.metrics "agent").metricsTuning.useDefaultAllowList (index .Values.metrics "agent").metricsTuning.useIntegrationAllowList }}
```
petewall, line 2:
```suggestion
{{ if or .Values.metrics.agent.metricsTuning.useDefaultAllowList .Values.metrics.agent.metricsTuning.useIntegrationAllowList }}
```

petewall: We only need the `(index .Values.metrics "node-exporter")` because that's how you reference anything with a dash in the key. So, in this case, we can go back to the standard method.

charts/k8s-monitoring/templates/agent_config/_agent.river.txt
```diff
@@ -1,7 +1,11 @@
...
 {{ "default_allow_lists/agent.yaml" | .Files.Get }}
 {{ end }}
+{{ if (index .Values.metrics "agent").metricsTuning.useIntegrationAllowList }}
```
petewall, line 5:
```suggestion
{{ if .Values.metrics.agent.metricsTuning.useIntegrationAllowList }}
```

petewall:
> the values file actually already had a .metrics.agent.metricsTuning.useIntegrationAllowList

Whoa! Looks like I originally had the same idea, but never finished the thought. Thanks for bringing this change.

petewall:
If the tests fail, try updating your branch (git fetch && git rebase origin/main) and try again.
Make sure to run `make regenerate-example-outputs` and check the diffs.

petewall (approved):

PR #377, opened 2024-02-13
Improve the docs, mostly by turing URLs into links.

petewall:
Also regenerate schema and other generated docs

Link to the README: https://github.com/grafana/k8s-monitoring-helm/tree/fix/improve-readme/charts/k8s-monitoring


petewall:
@cyrille-leclerc , this is an improvement, but the problem lies mostly in how [helm-docs](https://github.com/norwoodj/helm-docs) generates the values table. Unfortunately, it not much I can do while still using an auto-generated README.

skl (approved):

cyrille-leclerc:
Thanks Pete, makes sense.

PR #380, opened 2024-02-15
Improve the query test script so it can compare floating point numbers.

petewall:
Test cluster now has two nodes
Start checking for DPM
Also set the Loki ServiceMonitor to be 1m instead of the default 15s

### DIFF
--- a/charts/k8s-monitoring/ci/ci-2-values.yaml
+++ b/charts/k8s-monitoring/ci/ci-2-values.yaml
@@ -76,8 +76,8 @@ test:
   - query: "count(kube_pod_info{cluster=\"ci-test-cluster-2\", color=\"blue\"})"
     type: promql
     expect:
-      operator: ">"
       value: 1
+      operator: "<="  # There's gotta be at least one pod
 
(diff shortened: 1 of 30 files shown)


charts/k8s-monitoring/ci/ci-2-values.yaml
```diff
@@ -76,8 +76,8 @@ test:
   - query: "count(kube_pod_info{cluster=\"ci-test-cluster-2\", color=\"blue\"})"
     type: promql
     expect:
-      operator: ">"
       value: 1
+      operator: "<="  # There's gotta be at least one pod
```
skl, line 80: At least one, right?

```suggestion
      operator: ">="  # There's gotta be at least one pod
```

skl (approved):

charts/k8s-monitoring/ci/ci-2-values.yaml
petewall: The check is `<expected> <operator> <actual>` so this will be computed as `1 <= (actual pod count)`

PR #389, opened 2024-02-26
Add selectors to ServiceMonitor, PodMonitor and Probe objects

selyx:
Add label selector for PodMonitor, ServiceMonitor and Probe objects to limit the scraping to only certain targets.


petewall (approved):
Looks great! Thanks!

PR #401, opened 2024-03-01
Set extraMetricRelabelingRules correctly for all metric sources

petewall:


skl (approved):
lgtm - potential breaking change?

petewall:
> lgtm - potential breaking change?

Only if people were using extraMetricRelabelingRules and it now affects more metric sources than it was before.
Changing the incorrect extraRelabelingRules to extraMetricRelabelingRules shouldn't affect anything since the former will be working with discovery-related metrics and labels.

PR #403, opened 2024-03-05
Add the new hostinfo connector to traces pipeline

gouthamve:
This'll help users count the number of hosts producing traces.

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
@@ -47,8 +47,16 @@ otelcol.processor.k8sattributes "default" {
     logs    = [otelcol.processor.transform.add_resource_attributes.input]
 {{- end }}
 {{- if .Values.traces.enabled }}
-    traces  = [otelcol.processor.transform.add_resource_attributes.input]
+    traces  = [otelcol.processor.transform.add_resource_attributes.input, otelcol.connector.host_info.default.input]
 {{- end }}
+  }
+}
+
+otelcol.connector.host_info "default" {
+  // https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.connector.host_info/
+  host_identifiers = [ "k8s.node.name" ]
+  output {
+    metrics = [otelcol.processor.batch.batch_processor.input]
   }
 }
 
--- a/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
@@ -67,19 +67,20 @@ otelcol.connector.host_info "default" {
 
 otelcol.processor.batch "host_info_batch" {
   output {
+  {{- if or (eq .Values.externalServices.prometheus.protocol "otlp") (eq .Values.externalServices.prometheus.protocol "otlphttp") }}
+    metrics = [otelcol.processor.batch.metrics_service.input]
+  }
+}
+  {{- end }}
+  {{- if eq .Values.externalServices.prometheus.protocol "remote_write" }}
     metrics = [otelcol.exporter.prometheus.host_info_metrics.input]
   }
 }
 
 otelcol.exporter.prometheus "host_info_metrics" {
   add_metric_suffixes = false
-  {{- if eq .Values.externalServices.prometheus.protocol "remote_write" }}
-    forward_to = [prometheus.remote_write.metrics_service.receiver]
-  }
-  {{- end }}
-  {{- if or (eq .Values.externalServices.prometheus.protocol "otlp") (eq .Values.externalServices.prometheus.protocol "otlphttp") }}
-    forward_to = [otelcol.receiver.prometheus.metrics_service.receiver]
-  }
+  forward_to = [prometheus.remote_write.metrics_service.receiver]
+}
   {{- end }}
 {{- end }}
 
--- a/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
@@ -61,8 +61,17 @@ otelcol.processor.k8sattributes "default" {
 otelcol.connector.host_info "default" {
   host_identifiers = [ "k8s.node.name" ]
   output {
-    metrics = [otelcol.processor.batch.batch_processor.input]
+    metrics = [otelcol.processor.batch.host_info_batch.input]
   }
+}
+
+otelcol.processor.batch "host_info_batch" {
+  metrics = [otelcol.exporter.prometheus.host_info_metrics.input]
+}
+
+otelcol.exporter.prometheus "host_info_metrics" {
+  add_metric_suffixes = false
+  forward_to          = [prometheus.remote_write.metrics_service.receiver]
 }
 {{- end }}
 
--- a/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
@@ -47,18 +47,24 @@ otelcol.processor.k8sattributes "default" {
     logs    = [otelcol.processor.transform.add_resource_attributes.input]
 {{- end }}
 {{- if .Values.traces.enabled }}
-    traces  = [otelcol.processor.transform.add_resource_attributes.input, otelcol.connector.host_info.default.input]
+    traces  = [
+      otelcol.processor.transform.add_resource_attributes.input, 
+    {{- if .Values.metrics.enabled }}
+      otelcol.connector.host_info.default.input,
+    {{- end }}
+    ]
 {{- end }}
   }
 }
 
+{{- if and .Values.metrics.enabled .Values.traces.enabled }}
 otelcol.connector.host_info "default" {
-  // https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.connector.host_info/
   host_identifiers = [ "k8s.node.name" ]
   output {
     metrics = [otelcol.processor.batch.batch_processor.input]
   }
 }
+{{- end }}
 
 otelcol.processor.transform "add_resource_attributes" {
   // Grafana Cloud Kubernetes monitoring expects Loki labels `cluster`, `pod`, and `namespace`
--- a/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_processors.river.txt
@@ -66,13 +66,21 @@ otelcol.connector.host_info "default" {
 }
 
 otelcol.processor.batch "host_info_batch" {
-  metrics = [otelcol.exporter.prometheus.host_info_metrics.input]
+  output {
+    metrics = [otelcol.exporter.prometheus.host_info_metrics.input]
+  }
 }
 
 otelcol.exporter.prometheus "host_info_metrics" {
   add_metric_suffixes = false
-  forward_to          = [prometheus.remote_write.metrics_service.receiver]
-}
+  {{- if eq .Values.externalServices.prometheus.protocol "remote_write" }}
+    forward_to = [prometheus.remote_write.metrics_service.receiver]
+  }
+  {{- end }}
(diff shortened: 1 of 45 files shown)


jcreixell (approved):
LGTM, what happens when metrics are disabled? shall we detect that condition and trigger an error?

gouthamve:
> LGTM, what happens when metrics are disabled?

Excellent question. I just tested it and this is the error that is logged:
```
ts=2024-03-05T10:02:35.703316Z level=error msg="Error consuming metrics" component=otelcol.connector.host_info.default error="telemetry type is not supported"
```

>  shall we detect that condition and trigger an error?

Can you show me how can throw errors in helm charts? Don't work with helm much.

jcreixell:
not a helm expert (maybe @petewall @skl have feedback here), but there is https://helm.sh/docs/chart_template_guide/function_list/#fail

charts/k8s-monitoring/templates/agent_config/_processors.river.txt
```diff
@@ -47,11 +47,19 @@ otelcol.processor.k8sattributes "default" {
...
 {{- end }}
 {{- if .Values.traces.enabled }}
-    traces  = [otelcol.processor.transform.add_resource_attributes.input]
+    traces  = [otelcol.processor.transform.add_resource_attributes.input, otelcol.connector.host_info.default.input]
 {{- end }}
   }
 }
 
+otelcol.connector.host_info "default" {
+  // https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.connector.host_info/
```
petewall, line 56:
```suggestion
```
We don't need to include the documentation here.

petewall:
The output that leads to the new component and the component itself should be wrapped in something like:
```
{{- if .Values.metrics.enabled }}
...
{{- end }}
```

charts/k8s-monitoring/templates/agent_config/_processors.river.txt
```diff
@@ -47,11 +47,19 @@ otelcol.processor.k8sattributes "default" {
...
+  host_identifiers = [ "k8s.node.name" ]
```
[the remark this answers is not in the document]
gouthamve: The `k8sattributes` processor will set this.

gouthamve:
Good call @petewall, I've incorporated the changes, and this will now only be in effect when both metrics and traces are enabled.

petewall (approved):

petewall:
Do we have an example metric that is generated by this?
In this values file, we test that we received traces from an application on the cluster:
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/ci/ci-2-values.yaml#L94-L98
Can we add another PromQL query that ensures that we're getting the metric generated from this new component?

The "application" is actually another instance of Grafana Agent, not deployed by the Helm chart, but standalone. It's configuration is here:
https://github.com/grafana/k8s-monitoring-helm/blob/main/.github/configs/agent-config.yaml

gouthamve:
I'm running the tests and the agent is failing to write to Prometheus with:

```
ts=2024-03-05T15:20:49.38352509Z level=error msg="Exporting failed. The error is not retryable. Dropping data." component=otelcol.exporter.otlphttp.metrics_service error="Permanent error: error exporting items, request to https://prometheus-server.prometheus.svc:80/api/v1/otlp/v1/metrics responded with HTTP Status Code 400" dropped_items=2
ts=2024-03-05T15:21:05.403085459Z level=error msg="Exporting failed. The error is not retryable. Dropping data." component=otelcol.exporter.otlphttp.metrics_service error="Permanent error: error exporting items, request to https://prometheus-server.prometheus.svc:80/api/v1/otlp/v1/metrics responded with HTTP Status Code 400" dropped_items=2
ts=2024-03-05T15:21:11.407188962Z level=error msg="Exporting failed. The error is not retryable. Dropping data." component=otelcol.exporter.otlphttp.metrics_service error="Permanent error: error exporting items, request to https://prometheus-server.prometheus.svc:80/api/v1/otlp/v1/metrics responded with HTTP Status Code 400" dropped_items=611
ts=2024-03-05T15:21:15.402973422Z level=error msg="Exporting failed. The error is not retryable. Dropping data." component=otelcol.exporter.otlphttp.metrics_service error="Permanent error: error exporting items, request to https://prometheus-server.prometheus.svc:80/api/v1/otlp/v1/metrics responded with HTTP Status Code 400" dropped_items=299
```

Interesting.

gouthamve:
Fixed in the latest commit. This is because `k8sattributes` processor only adds the `k8s.node.name` attribute to traces coming from pods inside the k8s cluster. But the traces are produced by an agent outside the cluster. So I added it manually.

petewall:
Error from the CI test:
```
ts=2024-03-05T15:55:30.251470691Z level=error msg="failed to write target_info metric" component=otelcol.exporter.prometheus.metrics_converter err="1 error occurred:\n\t* job or instance cannot be found from labels\n\n"
ts=2024-03-05T15:55:30.251572862Z level=error msg="failed to write metric sample" component=otelcol.exporter.prometheus.metrics_converter traces_host_info=err "1 error occurred:\n\t* job or instance cannot be found from labels\n\n"=(MISSING)
```

charts/k8s-monitoring/templates/agent_config/_processors.river.txt
```diff
@@ -47,11 +47,25 @@ otelcol.processor.k8sattributes "default" {
...
 }
 
+{{- if and .Values.metrics.enabled .Values.traces.enabled }}
+otelcol.connector.host_info "default" {
+  host_identifiers = [ "k8s.node.name" ]
+  output {
+    metrics = [otelcol.processor.batch.batch_processor.input]
+  }
+}
+{{- end }}
```
rlankfo, line 67: This component should live closer to remote write. Something like this should allow us to skip the extra metric processing for the host info metrics:
```suggestion
    metrics = [otelcol.processor.batch.host_info_batch.input]
  }
}

otelcol.processor.batch "host_info_batch" {
  metrics = [otelcol.exporter.prometheus.host_info_metrics.input]
}

otelcol.exporter.prometheus "host_info_metrics" {
  add_metric_suffixes = false
  forward_to          = [prometheus.remote_write.metrics_service.receiver]
}

{{- end }}
```

charts/k8s-monitoring/templates/agent_config/_processors.river.txt
```diff
@@ -61,9 +61,18 @@ otelcol.processor.k8sattributes "default" {
...
   }
 }
+
+otelcol.processor.batch "host_info_batch" {
+  metrics = [otelcol.exporter.prometheus.host_info_metrics.input]
+}
+
+otelcol.exporter.prometheus "host_info_metrics" {
+  add_metric_suffixes = false
+  forward_to          = [prometheus.remote_write.metrics_service.receiver]
```
petewall, line 74:
```suggestion
  forward_to          = [prometheus.relabel.metrics_service.receiver]
```
Using this target instead of directly to the remote_write means that the `cluster` label will get set, and will properly use the desired protocol. (We don't always use remote_write for metrics)

gouthamve: We kinda don't want the cluster label set tbh. We will use this metric to bill users, and we bill them based on `count(traces_host_info)`, if there are more labels, there is a likelyhood that the same host is counted twice leading to overcounting.

While `cluster` label might not be causing any issue here we still want to minimise the number of label manipulations to make sure we send the simplest possible metric to Grafana Cloud. 

> We don't always use remote_write for metrics

Can you elaborate on this?

cc @rlankfo

petewall: Without the cluster label, it becomes more challenging to find out where the source of that metric is. Also, wouldn't you have problems differentiating between sources if you have, say, a dev and a prod cluster with identical deployments (and instrumented applications). The cluster label is what we use to know which cluster the metrics came from.

The `externalServices.prometheus` section allows you to change from `remote_write` to `otlp` or `otlphttp`:
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/values.yaml#L24

This was a request from a customer to be able to ship metrics via OTLP.

petewall: If the protocol is otlp, then the `prometheus.remote_write.metrics_service` component wont exist.
The `prometheus.relabel.metrics_service` component is the "fan-in" for all metric sources before being delivered for storage.

gouthamve: Okay I see what you mean:

```
Error: /etc/agent/config.river:109:26: component "prometheus.remote_write.metrics_service.receiver" does not exist or is out of scope

108 |   add_metric_suffixes = false
109 |   forward_to          = [prometheus.remote_write.metrics_service.receiver]
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
110 | }

Error: could not perform the initial load successfully
interrupt received
```

I got around it by copying some snippets. Please see my latest commit.

gouthamve: The error is from `otelcol.prometheus.receiver` and the following pipeline fails:

```
host_info_connector --> prometheus_exporter --> prometheus_receiver --> otlp output
```

Can we somehow have the following pipeline @petewall :

```
host_info_connector --> otlp output

OR

host_info_connector --> prometheus_exporter --> prometheus_remote_write
```

rlankfo: We'll have to do one of the two options presented by Goutham. In this case, we don't want to include any additional labels and the host info connector component should live as close to the remote write component as possible.

rlankfo: > Also, wouldn't you have problems differentiating between sources if you have, say, a dev and a prod cluster with identical deployments (and instrumented applications). The cluster label is what we use to know which cluster the metrics came from.

In this case, I don't think care to differentiate by cluster. If customers are sending data to a production cloud stack, we will charge them per host, regardless if that host is part of a dev cluster or not.

petewall: > > Also, wouldn't you have problems differentiating between sources if you have, say, a dev and a prod cluster with identical deployments (and instrumented applications). The cluster label is what we use to know which cluster the metrics came from.
> 
> In this case, I don't think we care to differentiate by cluster. If customers are sending data to a production cloud stack, we will charge them per host, regardless if that host is part of a dev cluster or not.

I guess my point is not that you necessarily want to differentiate between clusters, the point is to ensure that two different sources of metrics don't conflict. Node names are not guaranteed to be globally unique.

If you have two clusters, dev and prod, and each have nodes name "cluster-node-1", and "cluster-node-2", how do you know the difference between a metric tagged with "cluster-node-1" on dev and "cluster-node-1" on prod?

petewall: > Can we somehow have the following pipeline

Sure, just move the `{{- if ... }}` conditional that checks for the output protocol into the `forward_to` section of the `otelcol.connector.host_info` or `otelcol.processor.batch` components.

rlankfo: > If you have two clusters, dev and prod, and each have nodes name "cluster-node-1", and "cluster-node-2", how do you know the difference between a metric tagged with "cluster-node-1" on dev and "cluster-node-1" on prod?

this could certainly happen but I think it would be an edge case. Typically, the nodes will have different names when using AWS, GCP, Azure, etc.  In this case, we'd likely need the customer to fall back to `host.id` or some other host identifer resource attribute.

rlankfo: although thinking about it, adding the cluster label wouldn't increase the cardinality so we'd still get accurate host counts with our custom tracker so we can probably add it but we need to avoid adding any labels that will increase the cardinality.

gouthamve: The relabel processor can add more labels based on the config it's given tbh. It's not just cluster label, but I got around the issue for now and everything should be working now.

Could you take a look @petewall

petewall (approved):

PR #406, opened 2024-03-05
Add the ability for gathering pod logs via PodLogs objects

petewall:
Default to off, but it is tested with ci-2-values.yaml/


skl (approved):
Nice

jewbetcha (approved):

PR #426, opened 2024-03-20
Fix configuration format for filter processor overrides

aptomaKetil:
The metrics/logs/traces sections are blocks, not statements. Also, fix a missing closing bracket that caused the configuration to break if the user had configured any extra log filters.


petewall (approved):
Thank you for this! I really appreciate it!

PR #444, opened 2024-03-28
Export pod logs in OTLP format

mar4uk:
To make pod logs visible in App O11y they should have OTLP format.

Changing the format means that pod logs in kubernetes monitoring plugin will look like this:

Screenshot 2024-03-28 at 14 11 58

Which might not be the desired view for customers who don't want to see their pod logs in OTLP format

We can add a boolean parameter like `logs.pod_logs.exportInOTLPFormat` to the k8s-monitoring helm chart so the customer can choose format of logs


petewall:
I'm going to close this because we've updated the logs pipeline quite a bit since then. I believe all of this would work if you use the OTLP protocol in the loki external service.

PR #446, opened 2024-04-02
Add support for gathering profiles and delivering them to Pyroscope

petewall:
If `profiles.enabled=true`, then we'll deploy a Grafana Agent instance as a daemonset and it scrapes profiles from annotated pods as well as using ebpf.

Test code is updated to deploy OSS Pyroscope to the test cluster, and a query was added to one of the values file to check that profiles exists in the db.

Reviewer notes:
* No examples were touched except for the new `examples/profiles-enabled`. The only difference between that examlpe and default is the enabling of the profiles feature. You can ignore all example changes except maybe the output.yaml and profiles.river for the new example.
* The [values.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/446/files#diff-cd22d083c6f49784d198f55b990feb9ae245abf1353dd56a379ec880126f4488) has the most meaningful changes, since that's the "user interface" for this feature. Make sure it'd make sense for your users.
* Feel free to ignore `.github/configs/updatecli.d` It's the Helm chart's dependabot
* The generation of the agent config for profiles is split into 3 files:
  * Gathering and processing pprof profiles from annotations: https://github.com/grafana/k8s-monitoring-helm/pull/446/files#diff-a7d5d9e9827dc0a411c8c2a1853feb9eb2bdb5abf982e767b6505b84970dc0db (this was stolen from the agent config from the pyroscope helm chart)
  * Gathering and processing eBPF profiles: https://github.com/grafana/k8s-monitoring-helm/pull/446/files#diff-a864bb4baeae9e3b5795da963df14e43811503f6ecca62016db891121d2dd1b7 
  * Delivering profiles to Pyroscope: https://github.com/grafana/k8s-monitoring-helm/pull/446/files#diff-26b4fa3b600a3276456cf3753879c42270cc6e3ea97554b604c634ce5e819f48
*


scripts/setup-local-test-cluster.sh
```diff
@@ -62,6 +63,9 @@ helm upgrade --install loki-otlp grafana/grafana-agent -f "${GRAFANA_AGENT_LOKI_
 echo "Deploying Tempo..."
 helm upgrade --install tempo grafana/tempo -n tempo --create-namespace --wait
 
+echo "Deploying Pyroscope..."
+helm upgrade --install pyroscope grafana/pyroscope -f "${PYROSCOPE_VALUES}" -n pyroscope --create-namespace --wait
```
simonswine, line 67: By default this will not support multiple tenants. You will need to supply a extraArg like that:

```
--set pyroscope.extraArgs."auth.multitenancy-enabled"=true
```

simonswine (commented):
Generally looking good, will need to give a deeper look later.

.github/configs/grafana.yaml
```diff
@@ -27,3 +27,11 @@ datasources:
...
         type: tempo
         url: http://tempo.tempo.svc:3100
+
+      - name: Pyroscope
+        type: grafana-pyroscope-datasource
+        url: http://pyroscope.pyroscope.svc.cluster.local.:4040
+        jsonData:
+          httpHeaderName1: X-Scope-OrgID
+        secureJsonData:
+          httpHeaderValue1: "1"
```
petewall, line 37: Pre-configuring the test pyroscope data source in our test grafana instance.

.github/configs/pyroscope.yaml
```diff
@@ -0,0 +1,5 @@
+agent:
+  enabled: false
+pyroscope:
+  extraArgs:
+    "auth.multitenancy-enabled": true
```
petewall, line 5: This is how we're deploying the test pyroscope

charts/k8s-monitoring/test/query-test.sh
```diff
@@ -164,6 +180,11 @@ for i in $(seq 0 "${count}"); do
         exit 1
       fi
       ;;
+    profiles)
```
petewall, line 183: Does Pyroscope have a name for its query language? `ProfileQL`? `PyroQL`? The literal 🔥 emoji?

simonswine: So far we don't have a language other than the label selectors that we handle like LogQL/PromQL, but I think `profileql` or `pyroql` is good enough for those purposes. We can obvisouly also go on a UTF-8 journey with the emoji 😆

charts/k8s-monitoring/values.yaml
```diff
@@ -784,6 +828,43 @@ traces:
...
+  # TBD
+  # java:
+  #  enabled: true
+  #  namespaces: []
+  #  # -- Rule blocks to be added to the discovery.relabel component for Java profile sources. ([docs](https://grafana.com/docs/agent/latest/flow/reference/components/discovery.relabel/#rule-block))
+  #  extraRelabelingRules: ""
+
+
+  pprof:
+    # -- Gather profiles using pprof
```
simonswine, line 855:
```suggestion
    # -- Gather profiles by scraping pprof HTTP endpoints
```

charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
```diff
@@ -0,0 +1,63 @@
...
+  rule {
+    action = "replace"
+    source_labels = ["__meta_kubernetes_pod_container_name"]
+    target_label = "container"
+  }
+  // provide arbitrary service_name label, otherwise it will be set to {__meta_kubernetes_namespace}/{__meta_kubernetes_pod_container_name}
+  rule {
+    action = "replace"
+    regex = "(.*)@(.*)"
+    replacement = "ebpf/k8s/${1}/${2}"
```
simonswine, line 48: I do think it makes more sense without the `k8s/`  "folder".

```suggestion
    replacement = "ebpf/${1}/${2}"
```

charts/k8s-monitoring/values.yaml
```diff
@@ -784,6 +828,43 @@ traces:
...
+    namespaces: []
+    # -- Profile types to gather
+    types:
+      - memory
+      - cpu
+      - goroutine
+      - block
+      - mutex
+      - fgprof
+
```
simonswine, line 867: I do think `extraRelabelingRules` also make sense here

charts/k8s-monitoring/templates/agent_config/_profiles_service.river.txt
```diff
@@ -0,0 +1,35 @@
...
+{{- with .Values.externalServices.pyroscope }}
+pyroscope.write "profiles_service" {
+  endpoint {
+    url = nonsensitive(remote.kubernetes.secret.profiles_service.data[{{ .hostKey | quote }}])
+    headers = { "X-Scope-OrgID" = nonsensitive(remote.kubernetes.secret.profiles_service.data[{{ .tenantIdKey | quote }}]) }
+  }
+{{ if eq .authMode "basic" }}
+  basic_auth {
+    username = nonsensitive(remote.kubernetes.secret.metrics_service.data[{{ .basicAuth.usernameKey | quote }}])
+    password = remote.kubernetes.secret.metrics_service.data[{{ .basicAuth.passwordKey | quote }}]
```
simonswine, line 17: That seems like a copy paste problem 🙂:

```suggestion
    username = nonsensitive(remote.kubernetes.secret.profiles_service.data[{{ .basicAuth.usernameKey | quote }}])
    password = remote.kubernetes.secret.profiles_service.data[{{ .basicAuth.passwordKey | quote }}]
```

simonswine (approved):
Thank you for your work on this! I think this is almost there.

I found a few smaller issues, that need fixing, once they are fixed no concerns with merging. Almost LGTM 😆

charts/k8s-monitoring/templates/agent_config/_profiles_service.river.txt
petewall: Yes! Thanks for the catch!

charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
```diff
@@ -0,0 +1,63 @@
...
+  }
+{{- if .Values.profiles.ebpf.extraRelabelingRules }}
+{{ .Values.profiles.ebpf.extraRelabelingRules | indent 2 }}
+{{- end }}
+}
+
+pyroscope.ebpf "ebpf_pods" {
+  targets = discovery.relabel.ebpf_pods.output
+  forward_to = [pyroscope.write.profiles_service.receiver]
+}
```
simonswine, line 61: I think for this it would be good if we could expose `demangle` as a value, with the default value "none". In some cases it needs to be set to `"full"`.

```suggestion
  forward_to = [pyroscope.write.profiles_service.receiver]
  demangle = {{ .Values.profiles.ebpf.demangle }}"
}
```

charts/k8s-monitoring/templates/agent_config/_profiles_service.river.txt
```diff
@@ -0,0 +1,35 @@
...
+    password = remote.kubernetes.secret.profiles_service.data[{{ .basicAuth.passwordKey | quote }}]
+  }
+{{- end }}
+{{- if .tls }}
+  tls_config {
+  {{- range $k, $v := .tls }}
+    {{ $k }} = {{ $v | toJson }}
+  {{- end }}
+  }
+{{- end }}
```
simonswine, line 26: Basic auth needs to be under endpoint:

```suggestion
{{ if eq .authMode "basic" }}
    basic_auth {
      username = nonsensitive(remote.kubernetes.secret.profiles_service.data[{{ .basicAuth.usernameKey | quote }}])
      password = remote.kubernetes.secret.profiles_service.data[{{ .basicAuth.passwordKey | quote }}]
    }
{{- end }}
{{- if .tls }}
    tls_config {
    {{- range $k, $v := .tls }}
      {{ $k }} = {{ $v | toJson }}
    {{- end }}
    }
{{- end }}
  }
```

charts/k8s-monitoring/values.yaml
```diff
@@ -1213,6 +1297,35 @@ grafana-agent-logs:
...
+    configMap: {create: false}
+    clustering: {enabled: true}
+
+    securityContext:
+      privileged: true
+      runAsGroup: 0
+      runAsUser: 0
+
+  controller:
+    type: daemonset
```
simonswine, line 1318: This is important for eBPF to work

```suggestion
    type: daemonset
    hostPID: true
```

simonswine (commented):
Thanks for addressing so many things already, unfortunately a second review found some more 😆.

I think that's hopefully everything

charts/k8s-monitoring/templates/agent_config/_profiles_pprof.river.txt
```diff
@@ -0,0 +1,162 @@
...
+    target_label  = "container"
+  }
+{{- if .Values.profiles.pprof.extraRelabelingRules }}
+{{ .Values.profiles.pprof.extraRelabelingRules | indent 2 }}
+{{- end }}
+}
+
+{{- $profileTypes := .Values.profiles.pprof.types }}
+{{ range $profileTypes }}
+discovery.relabel "pprof_pods_{{.}}_default_name" {
```
skl, line 55: Just for my own understanding, what does this syntax mean: `{{.}}`

charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
```diff
@@ -0,0 +1,63 @@
...
+  }
+{{- end }}
+  role = "pod"
+}
+
+discovery.relabel "ebpf_pods" {
+  targets = discovery.kubernetes.ebpf_pods.targets
+  rule {
+    action = "drop"
+    regex = "Succeeded|Failed|Completed"
```
skl, line 21: Do you need `Pending` here as well? For consistency with `discovery.relabel.pprof_pods`

skl (approved):
🚀

charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
petewall: @simonswine ?

charts/k8s-monitoring/templates/agent_config/_profiles_pprof.river.txt
petewall: Typically. `.` is the top level data object. so `.Values` means delve inside the values `.Release` goes into Helm's pre-defined release object.

The `range` command iterates over a list and resets `.` to each item in that list. That's also why sometimes you see saving a variable before the range, as in with
```
$profileTypes := .Values.profiles.pprof.types
```
because that way we can still reference that item even though `.` has become something else.


You'll also see things like
```
{{ with .Values.externalServices.prometheus }}
```
which resets `.` to be starting at `.Values.externalServices.prometheus`.

petewall: In this case, `{{ . }}` is referencing the current element in the `.Values.profiles.pprof.types` array

charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
simonswine: No, having `Pending` ones included is fine for ebpf.

Although I noticed that according to https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase Completed is not a valid state

simonswine (approved):
Tested this e2e again! All looking good to me. Thanks for getting this together.

LGTM

Issue #459, opened 2024-04-15
Scraping pprof profiles: clustering=true + discovery.kubernetes selectors misconfiguration


Bruno-DaSilva:
I noticed after deploying profiling with pprof scraping enabled, my profiles were not properly being scraped. The daemonset pods that ran along side my applications were not showing the correct set of scrape targets, despite them being supplied as input.

I believe this is because of these two configurations. In particular, selecting ONLY the pods on the same node, combined with trying to spread them out with clustering enabled=true. If my reading of the docs is correct, clustering expects all daemonset pods to have the same input target list (not each agent pod having only the same workloads on the same node).

```
discovery.kubernetes "pprof_pods" {
      selectors {
        role = "pod"
        field = "spec.nodeName=" + env("HOSTNAME")
      }
      role = "pod"
    }
```

```
pyroscope.scrape "pyroscope_scrape_cpu" {
      targets = concat(discovery.relabel.pprof_pods_cpu_default_name.output, discovery.relabel.pprof_pods_cpu_custom_name.output)
      clustering {
        enabled = true
      }

      profiling_config {
        profile.process_cpu {
          enabled = true
        }
      }

      forward_to = [pyroscope.write.profiles_service.receiver]
    }
```

FYI I ran into this on the `feat/alloy` branch, if that makes a difference :) 

I believe the solution is to either set clustering enabled=false or remove the node selector. When I do so locally, scraping is now working.

cc @petewall (I see you added profiling 2 weeks ago in https://github.com/grafana/k8s-monitoring-helm/pull/446 and also working on `feat/alloy`)

petewall:
I'll remove clustering, since we're deployed as a daemonset anyway.

Bruno-DaSilva:
yay, thanks @petewall !

PR #461, opened 2024-04-15
Fix profile gathering when using a subset of pprof profile types

petewall:
Also disable clustering, since we pre-filter to pods on the same node.

### DIFF
--- a/charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
+++ b/charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
@@ -57,8 +57,10 @@ discovery.relabel "ebpf_pods" {
 
 pyroscope.ebpf "ebpf_pods" {
   targets = discovery.relabel.ebpf_pods.output
-  forward_to = [pyroscope.write.profiles_service.receiver]
+
   demangle = {{ .Values.profiles.ebpf.demangle | quote }}
+
+  forward_to = [pyroscope.write.profiles_service.receiver]
 }
 {{- end }}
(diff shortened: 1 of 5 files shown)


charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
```diff
@@ -57,8 +57,10 @@ discovery.relabel "ebpf_pods" {
 
 pyroscope.ebpf "ebpf_pods" {
   targets = discovery.relabel.ebpf_pods.output
-  forward_to = [pyroscope.write.profiles_service.receiver]
+
   demangle = {{ .Values.profiles.ebpf.demangle | quote }}
+
+  forward_to = [pyroscope.write.profiles_service.receiver]
```
simonswine, line 63: Nit: Not an improvement, most likely committed by accident.

simonswine (approved):
Thanks for fixing those bugs!

Fixes #461 #459 

LGTM

charts/k8s-monitoring/templates/agent_config/_profiles_ebpf.river.txt
petewall: This was more just my own style. When defining the fields inside a component, I like to do:
{
// inputs (where does the data come from)

// processing (what do you do with the data)

// delivery (where does the data go)
}

simonswine: I see and there it all makes sense 🙂

PR #463, opened 2024-04-16
Add more helper scripts to make it easier for other contributors and maintainers

petewall:
Also to do: Update the contributor guide to include:

For contributors, how to build, test, make prs
For maintainers, how to merge PRs, and release versions (update version, run the release workflow)

### DIFF
--- a/scripts/set-version.sh
+++ b/scripts/set-version.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+VERSION=$1
+
+if [ -z "${VERSION}" ] || [ "${VERSION}" == "-h" ] || [ "${VERSION}" == "--help" ]; then
+  echo "set-version.sh <version> [<appVersion>] - Update the chart version and regenerate generated content"
+  echo
+  echo "  <version> The version number to use for the chart."
+  echo "  <appVersion> The app version number to use for the chart (default to latest release)."
+fi
+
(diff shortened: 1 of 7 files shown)


scripts/set-version.sh
```diff
@@ -0,0 +1,21 @@
...
+VERSION=$1
+
+if [ -z "${VERSION}" ] || [ "${VERSION}" == "-h" ] || [ "${VERSION}" == "--help" ]; then
+  echo "set-version.sh <version> [<appVersion>] - Update the chart version and regenerate generated content"
+  echo
+  echo "  <version> The version number to use for the chart."
+  echo "  <appVersion> The app version number to use for the chart (default to latest release)."
+fi
+
+PLUGIN_VERSION=$(gh release list --repo grafana/grafana-k8s-plugin --limit 1 --json name --jq '.[].name')
```
skl, line 12: Might be worth noting this is a private repo

skl (approved):

scripts/set-version.sh
petewall: True. This would only be for maintainers, so Grafana Employees.
I want to make an override (didn't finish that part, I guess)
I'll also add a check before hand and a helpful error message.

jewbetcha (approved):

Issue #443, opened 2024-03-28
How to do tail sampling to drop health probes?


Lp-Francois:
Hello,

I tried in a [previous issue](https://github.com/grafana/k8s-monitoring-helm/issues/431#issuecomment-2020458505) to find a solution to my current problem: how to drop the traces coming from Kubernetes probes (readiness, liveness)? Sadly, using the span filter creates inconsistent traces, missing the root span as it drops the root.

Is there a way of doing tail sampling with this helm chart, in order to drop the full trace containing an health probe? (/health, /live or /ready)?

Or how would you recommend to proceed to drop these traces?

t00mas:
Hello François, I agree with you about moving to tail sampling to achieve your goal, instead of filtering.

Are you trying to get to something that looks like this?

```yaml
processors:
  tail_sampling:
    decision_wait: 10s
    num_traces: 100
    expected_new_traces_per_sec: 10
    policies:
      [
         {
            name: drop-probes-policy,
            type: string_attribute,
            string_attribute: {
              key: http.route,
              values: [\/livez, \/readyz, \/health],
              enabled_regex_matching: true,
              invert_match: true
            }
         }
      ]
```

Lp-Francois:
Hi @t00mas , exactly what I am trying to achieve :) 

Hence the PR https://github.com/grafana/k8s-monitoring-helm/pull/467 😁

rlankfo:
Please note we have another helm chart that you can use to include a tail sampling architecture: https://github.com/grafana/helm-charts/tree/main/charts/grafana-sampling

This chart deploys a layered set of agents to load balance and tail sample as described here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/README.md#scaling-collectors-with-the-tail-sampling-processor

This can be used in tandem with the k8s monitoring helm charts by exporting your traces to the load balancing layer of the sampling chart. This will give you the ability to scale the tail sampling architecture separately from the rest of your telemetry collection.

Lp-Francois:
Interesting @rlankfo,

Seems more complicated than just adding a stage to the config though :) 

But good to know for people that really needs tail sampling as it is a good work around

Lp-Francois:
[...]

Issue #431, opened 2024-03-22
Format issue with otelcol.processor.filter (traces.span)


Lp-Francois:
Hello!

I am currently trying to drop all health check probes from our traces as it generates a lot of noise (💸) and is not helpful.

For this I am configuring the values.yaml the following way:

```yaml
traces:
  enabled: true
  # Settings related to traces ingested via receivers
  receiver:
    # -- Apply a filter to traces received via the OTLP or OTLP HTTP receivers. ([docs](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.processor.filter/))
    filters:
      span:
        -  '"attributes[\"http.route\"] == \"/livez\"",'
        -  '"attributes[\"http.route\"] == \"/health\"",'
        -  '"attributes[\"http.route\"] == \"/readyz\"",'

# ...
```

which generates the following:

```hcl
// ...
otelcol.processor.filter "default" {
  error_mode = "ignore"
  traces = {
    span = [
        "attributes[\"http.route\"] == \"/livez\"",
        "attributes[\"http.route\"] == \"/readyz\"",
        "attributes[\"http.route\"] == \"/health\"",
    ]                                         <-- This is the line 145 highlighted by the error
  }

  output {
    metrics = [otelcol.processor.batch.batch_processor.input]
    logs = [otelcol.processor.batch.batch_processor.input]
    traces = [otelcol.processor.batch.batch_processor.input]
  }
}
// ...
```

The agent fails with:

```
Error: /etc/agent/config.river:145:6: missing ',' in field list
Error: could not perform the initial load successfully
Error: /etc/agent/config.river:145:6: missing ',' in field list
```

I reproduced in local with saving the generated config to a file `river` and ran:  `brew install grafana-agent && AGENT_MODE=flow grafana-agent fmt file.river`.

But I don't see any issue with the lines, in the generated files, other objects or arrays don't have comma and the agent doesn't complain 🤔 

Any idea what could be the issue?

Lp-Francois:
Wow that was fast! Thanks :)

petewall:
Yeah! Of course!
It's not in a release yet, but look for a new one in a little bit.

Lp-Francois:
OK will check on monday :) 

Have a nice weekend

petewall:
0.12.3 was just released! Let me know how it goes!

Lp-Francois:
[...]

PR #467, opened 2024-04-17
feat: add tail sampling for traces

Lp-Francois:
Add tail sampling for traces. Useful when dropping health probes from your traces for instance, as they can be quite noisy 💸.

Documentation: https://grafana.com/docs/grafana-cloud/monitor-applications/application-observability/setup/sampling/tail/

**Related issues:**
- https://github.com/grafana/k8s-monitoring-helm/issues/431
- https://github.com/grafana/k8s-monitoring-helm/issues/443

**Comments:**

- How to generate the example properly? I didn't find a way to generate the `examples/tail-sampling/*.alloy` files.
- Is there an easy way of testing the generated config? I made sure the helm chart output looks OK (with `helm template k8s-monitoring charts/k8s-monitoring -f examples/tail-sampling/values.yaml > examples/tail-sampling/output.yaml`), but didn't tested it would work correctly. 
  - If there is no official way, I guess I could copy paste from the helm chart output the generated config and pass it locally to the alloy cli?
- The allowed configuration is quite a lot: https://grafana.com/docs/alloy/latest/reference/components/otelcol.processor.tail_sampling/#blocks I wonder if there would be an easy way to pass config for the policies? Maybe like the `extraConfig` is working, just let the user pass any string?

**Note:** It is really painful to work with these `alloy_config/*.alloy.txt` files 😅


charts/k8s-monitoring/templates/alloy_config/_processors.alloy.txt
```diff
@@ -212,6 +212,40 @@ otelcol.processor.transform "default" {
...
+
+  {{- if and (eq $policy.type "string_attribute") (hasKey $policy "stringAttribute") }}
+    stringAttribute = {
+      key = {{ $policy.stringAttribute.key | quote }}
+      values = [
+      {{- range $value := $policy.stringAttribute.values }}
+        {{ $value }},
+      {{- end }}
+      ]
+      enabled_regex_matching = {{ $policy.stringAttribute.enabledRegexMatching | default true }}
```
t00mas, line 236: I think both `enabled_regex_matching` and `invert_match` can be wrapped in a {{- if }} {{- end }} block, since they're not required in the `string_attribute` block spec.

Also, `enabled_regex_matching` should default to `false`.

charts/k8s-monitoring/templates/alloy_config/_processors.alloy.txt
```diff
@@ -212,6 +212,45 @@ otelcol.processor.transform "default" {
   }
 }
 
+
+{{- if and .Values.traces.enabled .Values.traces.receiver.tailsampling.policies }}
```
t00mas, line 216: Also, this errors when running `make regenerate-example-outputs`

`nil pointer evaluating interface {}.policies`

at `traces-enabled/values.yaml`

It is possible to have a `nil` `tailsampling` section, so a stricter check is necessary for this.

Lp-Francois:
@t00mas I applied your reviews :) 

Could you have another look if you find some time please 🙏 ?

t00mas:
I'm OK with this, just take into account @rlankfo 's recommendation here https://github.com/grafana/k8s-monitoring-helm/issues/443#issuecomment-2096730140

t00mas (approved):

rlankfo:
I'd be concerned about adding the tail sampling processor at this stage of the pipeline. It won't be scalable -- to scale the tail sampling processor we recommend introducing a load balancing exporter in a separate layer that can route traces by trace ID.  @petewall do you have any thoughts here?

a-abella:
I am hitting the need for this now and have some issues with the solution proposed in https://github.com/grafana/k8s-monitoring-helm/issues/443#issuecomment-2096730140. It doesn't integrate well with the k8s-monitoring chart because k8s-monitoring supports a variety of authentication mechanisms for different backends that the [grafana-sampling](https://github.com/grafana/helm-charts/tree/main/charts/grafana-sampling) chart does not. Additionally the grafana-sampling chart has no built-in support for OTTL metric/datapoint filtering like this chart does. 

If sampling within this k8s-monitoring chart is not encouraged due to scalability concerns then instead of including sampling  templates, a suitable alternative could be a way to add an alloy `extraConfig` pre-batch processor within the `traces` template, allowing users to configure their own sampling in a way that integrates with this chart at their own peril.

Alternatively, and I have no idea how possible this is, it would be interesting if this chart and the grafana-sampling chart could interop such that spans and OTEL metrics arrive at the k8s-monitoring statefulset, Traces are forwarded to grafana-sampling load balancing endpoint for processing, then the grafana-sampling configuration points its outputs back to k8s-monitoring which handles final output/export to backends. This would allow k8s-monitoring to handle the routing of metric, log, and trace data to its various supported backend datasources while still providing a tail sampler that can scale with load.

PR #471, opened 2024-04-17
Create project assignment workflow

petewall:
Notice the project assignment. The workflow did that... not me!


petewall:
How it looks when not logged in:
image

skl (approved):

PR #474, opened 2024-04-17
Turn off clustering for log gathering alloy

petewall:
Move PodLogs control to the primary alloy, which does use clustering.

Also add some instructions for setting the storage path for logs instance.

### DIFF
--- a/charts/k8s-monitoring/templates/_config_validations.tpl
+++ b/charts/k8s-monitoring/templates/_config_validations.tpl
@@ -2,15 +2,54 @@
 {{- if and .Values.logs.enabled .Values.logs.pod_logs.enabled }}
   {{- if eq .Values.logs.pod_logs.gatherMethod "volumes" }}
     {{- if ne (index .Values "alloy-logs").controller.type "daemonset" }}
+{{/*
+Invalid configuration for gathering pod logs! When using logs.pod_logs.gatherMethod: "volumes", Grafana Alloy for Logs must be a Daemonset. Otherwise, logs will be missing!
+Please set:
+logs:
+  pod_logs:
+    gatherMethod: api
+or
+alloy-logs:
+  controller:
+    type: daemonset
+*/}}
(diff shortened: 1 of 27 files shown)


charts/k8s-monitoring/templates/_config_validations.tpl
```diff
@@ -2,15 +2,54 @@
...
+  alloy:
+    clustering:
+      enabled: false
+*/}}
+      {{ fail "Invalid configuration for gathering pod logs! When using logs.pod_logs.gatherMethod: \"volumes\", Grafana Alloy for Logs should not utilize clustering. Otherwise, performance will suffer!\nPlease set:\nalloy-logs:\n  alloy:\n    clustering:\n      enabled: false"}}
+    {{- end }}
+
   {{- else if eq .Values.logs.pod_logs.gatherMethod "api" }}
     {{- if not (index .Values "alloy-logs").alloy.clustering.enabled }}
       {{- if eq (index .Values "alloy-logs").controller.type "daemonset" }}
```
thampiotr, line 32: Logs via k8s API can be collected with a deployment or statefulset, a daemonset ties the number of instances to number of nodes, which is not necessary. Technically this is not wrong though.

examples/alloy-autoscaling-and-storage/values.yaml
```diff
@@ -42,4 +42,24 @@ alloy:
...
+    enableStatefulSetAutoDeletePVC: true
+    volumeClaimTemplates:
+      - metadata:
+          name: alloy-log-positions
+        spec:
+          accessModes: ["ReadWriteOnce"]
+          storageClassName: "standard"
+          resources:
+            requests:
+              storage: 100Mi
```
thampiotr, line 65: Looks good, just that we haven't been running it this way, so not much experience on our side.

thampiotr (approved):
Thanks!

skl (approved):

PR #481, opened 2024-04-19
chore: fix indentation for jaeger component

### LINKED ISSUE #477: Jaeger component is not indented correctly
I don't think this is functionally a problem, but it we should fix it:
https://github.com/grafana/k8s-monitoring-helm/blob/main/examples/traces-enabled/metrics.alloy#L37-L52

jewbetcha:
Real big change here

Fixes #477


petewall (approved):
🎉

PR #482, opened 2024-04-19
Make the host info metric generation optional

petewall:
Making this optional, because the application observability plugin in Grafana Cloud will bill for these metrics.


petewall:
Converting to draft. I want to create an "Application observability" example that:
1. Enables traces with this feature turned on
2. Enables profiles and pyroscope
3. Describes gathering data from Beyla

petewall:
Also, what do we do about AppO11y customers that were using this previously, then upgrade? They'll lose AppO11y features, making this a potentially breaking change.

charts/k8s-monitoring/values.schema.json
```diff
@@ -1297,24 +1303,6 @@
...
-                                    }
-                                },
-                                "includeMetrics": {
-                                    "items": {
-                                        "type": "string"
-                                    }
-                                }
-                            }
-                        }
-                    }
```
petewall, line 1317: This was apparently a duplicate `alloy` section

charts/k8s-monitoring/templates/alloy_config/_processors.alloy.txt
```diff
@@ -69,15 +69,15 @@ otelcol.processor.k8sattributes "default" {
 {{- if .Values.traces.enabled }}
     traces  = [
       otelcol.processor.transform.default.input, 
-    {{- if .Values.metrics.enabled }}
+    {{- if and .Values.metrics.enabled .Values.traces.hostInfoMetrics.enabled }}
```
jcreixell, line 72: The `otelcol.connector.host_info` component will likely be renamed to `otelcol.connector.grafanacloud`, following its OTEL collector contrib [upstream name](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/grafanacloudconnector).

I would propose to call this flag `grafanaCloudMetrics` instead of `hostInfoMetrics`.

@rlankfo WDYT?

jcreixell:
> Also, what do we do about AppO11y customers that were using this previously, then upgrade? They'll lose AppO11y features, making this a potentially breaking change.

I'd propose the following:

1. Set the default value of the new flag to `true`
2. Rollout plugin changes that allow disabling this functionality
3. Set the default value to false in the chart (with the plugin still defaulting to `true`)

WDYT?

charts/k8s-monitoring/templates/alloy_config/_processors.alloy.txt
rlankfo: This sounds good to me!

petewall:
Ready to review, please take a look and let's get the process going.
When this is merged and released, I'll do an associated change to make this disabled in the plugin.

charts/k8s-monitoring/README.md
```diff
@@ -724,6 +724,7 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | traces.enabled | bool | `false` | Receive and forward traces. |
+| traces.grafanaCloudMetrics.enabled | bool | `true` | Generate host info metrics from trace data, used in Application Observability in Grafana Cloud. |
```
jcreixell, line 727:
```suggestion
| traces.grafanaCloudMetrics.enabled | bool | `true` | Generate host info metrics from telemetry data, used in Application Observability in Grafana Cloud. |
```

rationale: the connector is likely to incorporate support for other telemetry signals (like OTEL HTTP) in the future

jcreixell (commented):
overall it LGTM. However, I think it would be idea to try to disassociate the grafanaCloudConnector from traces (as in `traces.grafanaCloudMetrics.enabled`), since we expect to generate this metric from other signals types in the future

charts/k8s-monitoring/README.md
petewall: Would it make more sense to move this outside of the "traces" section in the values file and inside the "receivers" section? That's where we've been putting things that apply to all telemetry data gathered via otlp.

charts/k8s-monitoring/README.md
```diff
@@ -646,6 +646,18 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely
 |-----|------|---------|-------------|
 | receivers.deployGrafanaAgentService | bool | `true` | Deploy a service named for Grafana Agent that matches the Alloy service. This is useful for applications that are configured to send telemetry to a service named "grafana-agent" and not yet updated to send to "alloy". |
 
+### OTEL Receivers (Processors)
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| receivers.grafanaCloudMetrics.enabled | bool | `true` | Generate host info metrics from trace data, used in Application Observability in Grafana Cloud. |
```
jcreixell, line 653:
```suggestion
| receivers.grafanaCloudMetrics.enabled | bool | `true` | Generate host info metrics from telemetry data, used in Application Observability in Grafana Cloud. |
```

charts/k8s-monitoring/values.yaml
```diff
@@ -1610,6 +1610,10 @@ receivers:
       # @section -- OTEL Receivers (Processors)
       annotations: []
 
+  grafanaCloudMetrics:
+    # -- Generate host info metrics from trace data, used in Application Observability in Grafana Cloud.
```
jcreixell, line 1614:
```suggestion
    # -- Generate host info metrics from telemetry data, used in Application Observability in Grafana Cloud.
```

jcreixell (approved):
2 small nits, otherwise LGTM!

PR #487, opened 2024-04-24
Update OpenShift platform and examples

petewall:
This creates three security context constraints objects now, one for each Alloy instance. The code does template once and invokes that template three times (the logs template will enable hostpath permissions if it's using volume based log gathering)

It does not set the user id to non-root.

It does add a set of capabilities that was listed in another issue. At some point, I'll need to sync with the alloy team for the minimal set required here.

### DIFF
--- a/examples/openshift-compatible/README.md
+++ b/examples/openshift-compatible/README.md
@@ -1,5 +1,7 @@
 # OpenShift Compatible
 
+TODO THESE PARAGRAPHS NEED UPDATING
+
 This example shows the modifications from the [default](../default-values) to deploy Kubernetes Monitoring on an OpenShift cluster.
 
 These modifications prevent deploying Kube State Metrics and Node Exporter, since they will already be present on the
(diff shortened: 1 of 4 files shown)


examples/openshift-compatible/README.md
```diff
@@ -52,22 +54,52 @@ prometheus-node-exporter:
 
 alloy:
   alloy:
-    listenPort: 8080
     securityContext:
+      runAsUser: 473
+      runAsGroup: 473
```
rfratto, line 59: I think this needs to come with a huge warning that it would be using undocumented and unsupported functionality (essentially experimental). We don't know anything about the future of the non-root user in the Alloy container, including the user/group IDs.

petewall: Re-rolled the example without the user assignment.

examples/openshift-compatible/README.md
```diff
@@ -1,13 +1,15 @@
 # OpenShift Compatible
 
+TODO THESE PARAGRAPHS NEED UPDATING
```
skl, line 3: Still needed?

skl (approved):
lgtm - out of interest, where does this change come from or what's the issue being resolved?

examples/openshift-compatible/README.md
petewall: Ahh, geez.. The ironic thing is that I put that there to remind myself to deal with that before comitting.
I need to make a pre-commit hook just for myself...

petewall:
This comes from an escalation: https://github.com/grafana/support-escalations/issues/10230

PR #497, opened 2024-04-29
Dump the generated alloy config when the helm validate hook fails.

HackToHell:
Quality of life improvement, it's hard to dig out the file from the configmap otherwise.

### DIFF
--- a/charts/k8s-monitoring/templates/hooks/validate-configuration.yaml
+++ b/charts/k8s-monitoring/templates/hooks/validate-configuration.yaml
@@ -47,6 +47,8 @@ spec:
       - |
         echo Validating Grafana Alloy config file
         if ! alloy fmt /etc/alloy/config.alloy > /dev/null; then
+          echo Generated Config with Errors
+          cat /etc/alloy/config.alloy
           exit 1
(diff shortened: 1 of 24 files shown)


charts/k8s-monitoring/templates/hooks/validate-configuration.yaml
```diff
@@ -47,6 +47,8 @@ spec:
       - |
         echo Validating Grafana Alloy config file
         if ! alloy fmt /etc/alloy/config.alloy > /dev/null; then
+          echo Generated Config with Errors
```
petewall, line 50: Can we change this message to say something more like "Grafana Alloy config file has syntax errors"
Similarly, after the "alloy run" check, and before echoing the output, can you add "Grafana Alloy config file has errors"?

petewall:
Coming back to this, I don't think this is a great idea because the generated configs will be just under a 1000 lines long. It'll easily flood a terminal and cause lots of back-scrolling just to find the error message in the logs.

When it fails, it typically prints the lines that were in error, so it should point to exactly where to look in the configmap.

PR #525, opened 2024-05-15
Make the Grafana Agent compatibility service optional

petewall:


jewbetcha (approved):
LGTM. Only concern would be that if users use this configuration right now, will they ever actually fully switch?

petewall:
> LGTM. Only concern would be that if users use this configuration right now, will they ever actually fully switch?

It's already on, this just makes it optional. Eventually, we can switch this to `false` and print a deprecation warning.

PR #527, opened 2024-05-15
Prometheus Operator Relabelings

### LINKED ISSUE #526: Add Support for Service Discovery Relabelings for Prometheus Operator Objects

bentonam:
The Alloy components: 

- `prometheus.operator.podmonitors`
- `prometheus.operator.probes`
- `prometheus.operator.servicemonitors`

All support discovery relabeling rules, but instead of coming from the `discovery.relabel` component, the rules are embedded inside each component and are ran after the objects `relabelings` have been applied.  

Resolves #526


examples/custom-prometheus-operator-rules/README.md
```diff
@@ -0,0 +1,67 @@
+# Custom Prometheus Operator Rules
```
petewall, line 1: Nice! Can you add a link to this in `examples/README.md`?

PR #528, opened 2024-05-15
Add the ability to extract labels and annotation in the k8sattributes processor

### LINKED ISSUE #395: Customizable Loki Resource Labels
Hi, the loki filter fields is coming as default and we cannot customize it to add new fields.

image

I tried to update but no dynamic config parameter in here. Only exists `{{- if .Values.logs.enabled }}` and then it adds all scope staticly.
https://github.com/grafana/k8s-monitoring-helm/blob/df734dde136962477ec398fdcfcc3ad998a0bc3a/charts/k8s-monitoring/templates/agent_config/_processors.river.txt#L72

How can I add new filter fields to Loki results? Is there another option or one way is changing this `attributes[\"loki.resource.labels\"]` value?

petewall:
Also add the ability to customize the log labels set for loki.

Fixes #395

### DIFF
--- a/.github/configs/.gitleaks.toml
+++ b/.github/configs/.gitleaks.toml
@@ -1,0 +1,3 @@
+[allowlist]
+description = "Ignore output.yaml files in examples directory"
+path = '''examples/.*/output.yaml'''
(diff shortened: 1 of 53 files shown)


rlankfo (approved):
LGTM! Thanks Pete 💯

PR #530, opened 2024-05-15
Add Kepler to gather energy usage data

petewall:
This PR adds Kepler as another subchart dependency. This means that we can optionally deploy Kepler and tell Alloy to scrape metrics from it.

Notes to reviewers:
* The values.yaml file is the main interface for the helm chart.
  * The `metrics.kepler` section controls how we configure alloy to scrape kepler
  * The `kepler` section controls how kepler is deployed.
* `.github` are for actions, mostly around the tooling that keeps kepler helm chart up to date.
* `charts/k8s-monitoring/ci/*.values.yaml` are the test files for our CI tooling. Kepler is enabled in one of those files and we add a check for a known metric to ensure that end-to-end delivery is working
* anything inside of `examples` are automatically generated


petewall:
Ignore the failing `markdownlint` error. That's a known problem with the data in the README.

skl (approved):
🌍

charts/k8s-monitoring/values.yaml
```diff
@@ -2029,6 +2080,18 @@ opencost:
...
+# See https://github.com/sustainable-computing-io/kepler-helm-chart/tree/main/chart/kepler for available values.
+kepler:
+  # -- Should this Helm chart deploy Kepler to the cluster.
+  # Set this to false if your cluster already has Kepler, or if you do not want to scrape metrics from Kepler.
+  # @section -- Chart
+  enabled: false
+
+  # @ignored -- This skips including these values in README.md
+  canMount:
+    usrSrc: false
```
jjo, line 2094: `values.yaml` does have a `nodeSelector` field defaulting to
```
nodeSelector:
  kubernetes.io/os: linux
```
, suggest adding also amd64 constraint, as per https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-arch
```
  kubernetes.io/arch: amd64
```

jjo, line 2094: Also consider adding below to `extraEnvVars:`

```
  EXPOSE_ESTIMATED_IDLE_POWER_METRICS: "true"
```
this was recommended from #kepler-project channel at CNCF slack

petewall: Yeah, I really thought about adding this. My hesitation is that there are ARM platforms that are supported, and I don't want to cut them out.

Is it easier to have the amd64 restriction and have to explain how to remove if you have a supported arm node, or don't have the restriction and have a doc about why pods are crashing on graviton nodes. I'm leaning towards the latter, but happy to discuss.

petewall: I looked a long time to determine if I can detect if the pod is on a graviton node, but I don't think it's possible at pod scheduling time.

jjo (approved):
LGTM, this is greatO

PR #531, opened 2024-05-16
Added support for Helm Doc Sections

bentonam:


petewall (approved):
This is great! I never knew it could do this.

PR #543, opened 2024-05-21
Alloy Metric Module Imports

bentonam:


petewall (commented):
I love making modules become a first-class citizen.
I want to take some time to properly review and test this, so it'll likely be next week when I get around to it.

petewall (approved):
I think this is looking good.
Is there a doc in alloy_modules that describes the convention that we're starting to enforce here?

bentonam:
> I think this is looking good. Is there a doc in alloy_modules that describes the convention that we're starting to enforce here?

@petewall I don't think the is a doc that officially describes it, the convention is implemented in the majority of metric modules after some discussions with multiple people from different teams.  It's hard to say if we should create one or not, because it's not like a model "has to" implement that, but for use within the helm chart we need to have consistency.   If a user doesn't want that, they can still do it through the `extraConfig`.  

I updated the `module.alias` to `module.name`, merged in changes from `main` and updated the `values.schema.json`.  Do you want to review before merging?

petewall (approved):

PR #545, opened 2024-05-22
Added Support for Journal Logs

### LINKED ISSUE #544: Add Support for Worker Journal Logs

bentonam:
Resolves #544


petewall (commented):
Looks good! Does this work with Kind clusters? Can you add a check on one of our ci/*values.yaml files to include this and add a logql query to verify we get the data?

bentonam:
@petewall Kind nodes do not have systemd.  I think it might be possible to mount the container logs to `/var/log/journal` but they would be in docker json format and not journal format.  We would probably have to ingest those logs with fluentd or something like that, covert them and write them back out, it might work but I'm not certain of that and it seems like a lot of extra complexity.  

Would it be better to spin up a cluster in GCP or AWS to test as part of CI/CD?  That is obviously a bigger task, but might be worthwhile as the chart and k8s integration can begin to collect different information depending on the provider.

petewall (approved):

PR #552, opened 2024-06-03
Dynamic External Labels

bentonam:
Allow external labels to be added unquoted which allows for dynamic references.  

```yaml
  loki:
    host: https://loki.example.com
    basicAuth:
      username: 12345
      password: "It's a secret to everyone"
    externalLabels:
      region: southwest
      tenant: widgetco
    externalLabelsFrom:
      env: remote.kubernetes.configmap.cluster_info.data["env"]
      region: remote.kubernetes.configmap.cluster_info.data["region"]
      provider: env("PROVIDER")
```


rosariofe (approved):

examples/custom-config/values.yaml
```diff
@@ -11,6 +11,9 @@ externalServices:
     externalLabels:
       region: southwest
       tenant: widgetco
+    externalLabelsFrom:
+      env: remote.kubernetes.configmap.cluster_info.data["env"]
+      region: remote.kubernetes.configmap.cluster_info.data["region"]
```
petewall, line 16: Can you update this example's README.md with the updated values file? Eventually, I want a script that syncs the two, or at least warns if they're out of sync.

petewall (commented):
Can you add one or more of these to the values files in the ci directory and add a check in the test queries? Like, add something like:
`MAGIC_WORD: abracadabra`
As an env var on the alloy instance. And then use the new thing to set that as a label from the var, then add that to one or more of the test queries.

petewall (approved):

PR #577, opened 2024-06-17
fix(docs): pluralization in referenced keys

nferch:


petewall:
Thank you for your changes! Great catch.
`README.md` is actually generated from the chart itself. Can you actually make these changes in `charts/k8s-monitoring/values.yaml` and then run `helm-docs` (https://github.com/grafana/k8s-monitoring-helm/blob/main/CONTRIBUTING.md#building-generated-files)?

nferch:
> Thank you for your changes! Great catch. `README.md` is actually generated from the chart itself. Can you actually make these changes in `charts/k8s-monitoring/values.yaml` and then run `helm-docs` (https://github.com/grafana/k8s-monitoring-helm/blob/main/CONTRIBUTING.md#building-generated-files)?

Also had to update `charts/k8s-monitoring/README.md.gotmpl`, but it looks that running `make README.md` in `charts/k8s-monitoring` should take care of that as well.

petewall (approved):

PR #579, opened 2024-06-18
Add TLS blocks to each receiver that supports it

petewall:


examples/traces-enabled/values.yaml
```diff
@@ -47,3 +51,16 @@ receivers:
       enabled: true
   zipkin:
     enabled: true
+    tls:
+      ca_pem: |
+        -----BEGIN CERTIFICATE-----
+        Example CA certificate
```
skl, line 57: Seems legit

skl (approved):

PR #581, opened 2024-06-20
Added support to modify the scrape_interval for Auto-Discovery jobs

bradleypettit:
This adds support for configuring the scrapeInterval for the Auto-Discovery jobs (http & https) in the same way it can be modified for all other scrape jobs.


bradleypettit:
Closed and replaced with https://github.com/grafana/k8s-monitoring-helm/pull/590.

PR #583, opened 2024-06-20
feat: Add network metrics to node exporter allow list

skl:
Adds network metrics to node exporter default allow list.


jewbetcha (approved):

petewall (approved):
LGTM

PR #590, opened 2024-06-24
Added support to modify the scrape_interval for Auto-Discovery jobs

bradleypettit:
This adds support for configuring the scrapeInterval for the Auto-Discovery jobs (http & https) in the same way it can be modified for all other scrape jobs.


petewall:
I think this is great! I think it would be a great addition to also support a `k8s.grafana.com/metrics.scrapeInterval` annotation  to set the `__scrape_interval__` label. What do you think about adding that, too?

bradleypettit:
> I think this is great! I think it would be a great addition to also support a `k8s.grafana.com/metrics.scrapeInterval` annotation to set the `__scrape_interval__` label. What do you think about adding that, too?

Great idea Pete, added that functionality as well.

petewall (approved):

petewall:
I think as it stands now, it'll always overwrite the annotation value. I'm not sure if I want the annotation to take precedence over the helm chart values, though. Let me think about this for a bit.

bradleypettit:
Tested the behaviour in my lab environment now. Currently the values file will set the default scrape_interval, and annotations (if set) will override that scrape_interval.

IMO this is ideal as I can have a lower DPM (60s) by default in my cluster, and selectively set a more aggressive scrape_interval (15s) for more critical workloads via their annotations.

petewall:
Oh awesome. I was concerned that setting the scrape interval in the prometheus.scrape component was going to override the __scrape_interval__ meta label. Sounds like it's the opposite. Great!

bradleypettit:
Yeah TIL as well! Thanks Pete.

PR #600, opened 2024-06-27
fix: allow for receiver service name override

jdcmarques:
## Description
We want to be able to pick the name of the receiver service, but after the migration to alloy whenever we pass in
`.Values.grafana-agent.fullnameOverride` we get the following error

```bash
Error: execution error at (grafana-chart/charts/k8s-monitoring/templates/kubernetes-monitoring-telemetry.yaml:1:3): 

As of k8s-monitoring Chart version 1.0, Grafana Agent has been replaced with Grafana Alloy.
These sections in your values file will need to be renamed:
  grafana-agent          --> alloy
  grafana-agent-events   --> alloy-events
  grafana-agent-logs     --> alloy-logs
  grafana-agent-profiles --> alloy-profiles
  metrics.agent          --> metrics.alloy

For more information, see https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring#breaking-change-announcements

```

### DIFF
--- a/charts/k8s-monitoring/templates/_helpers.tpl
+++ b/charts/k8s-monitoring/templates/_helpers.tpl
@@ -96,11 +96,11 @@
 {{ . | replace "-" "_" | replace "." "_" | replace "/" "_" }}
 {{- end }}
 
-{{- define "grafana-agent.fullname" -}}
-{{- if (index .Values "grafana-agent").fullnameOverride }}
-{{- (index .Values "grafana-agent").fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- define "alloy.fullname" -}}
+{{- if (index .Values "alloy").fullnameOverride }}
+{{- (index .Values "alloy").fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
-{{- $name := default "grafana-agent" (index .Values "grafana-agent").nameOverride }}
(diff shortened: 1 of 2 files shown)


petewall:
Yeah, I agree that this is a problem, but changing this to use `alloy` will not really solve it.

When we released version 1.0, we moved from Grafana Agent to Alloy and added that deprecation warning. I then added a service that deploys using the existing Grafana Agent naming, with the idea that if you had applications configured to send data via OTLP to the agent, it wouldn't break those applications. It makes a duplicated service.

If you really want to pick the name of the service, you have a few options:
1. Use `alloy.fullnameOverride` to rename Alloy itself. The associated service will be renamed.
2. Use the `extraObjects` to deploy a service that points to Alloy.

jdcmarques:
Thanks for clearing that up!

I will double-check why the alloy service is not getting created when deploying with ArgoCD https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/service.yaml as that would allow for this to work by default. This is most likely due to having `deployGrafanaAgentService` as true by default

Can update this to look for a `.Values.receivers.grafanaAgentService:` object with `fullNameOverride` or `nameOverride`, would be right next to the value `deployGrafanaAgentService` that enables this.

petewall:
In v2, we just added the ability to create a service that will map to the `alloy-receivers` instance and you can pick the name of that service now.

PR #607, opened 2024-07-01
Alloy Integration: edit allow list

gaantunes:
This is the correct list of metrics to be used together with the [alloy mixin](https://github.com/grafana/alloy/tree/main/operations/alloy-mixin).

Let me know if you want to keep any of the other existing metrics I have deleted, but they won't be used with the integration.


charts/k8s-monitoring/default_allow_lists/alloy_integration.yaml
```diff
@@ -1,57 +1,65 @@
 ---
 # The set of metrics from Grafana Alloy required for the Grafana Alloy integration
-- alloy_tcp_connections
-- alloy_wal_samples_appended_total
-- alloy_wal_storage_active_series
+- alloy_build_info
```
skl, line 3: `alloy_build_info` is already in by default so not needed here, although it doesn't do any harm either way.

```suggestion
```

skl (commented):
@petewall I'll leave approval for you as there's a bunch of metrics removed here - maybe they need separating out to another file if they're captured for purposes other than the alloy integration?

charts/k8s-monitoring/default_allow_lists/alloy_integration.yaml
petewall: I'm fine with duplicating. This protects against things like:
```
metricsTuning:
  useDefaultAllowList: false
  useIntegrationAllowList: true
```

PR #626, opened 2024-07-08
Add ability to set Loki labels from Pod labels

petewall:
Chose to set `app_kubernetes_io_name` from the Pod label `app.kubernetes.io/name`, because it's the most clear what is happening.


charts/k8s-monitoring/README.md
```diff
@@ -400,6 +400,7 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely
...
 | logs.pod_logs.extraStageBlocks | string | `""` | Stage blocks to be added to the loki.process component for pod logs. ([docs](https://grafana.com/docs/alloy/latest/reference/components/loki.process/#blocks)) This value is templated so that you can refer to other values from this file. |
 | logs.pod_logs.gatherMethod | string | `"volumes"` | Controls the behavior of gathering pod logs. When set to `volumes`, Grafana Alloy will use HostPath volume mounts on the cluster nodes to access the pod log files directly. When set to `api`, Grafana Alloy will access pod logs via the API server. This method may be preferable if your cluster prevents DaemonSets, HostPath volume mounts, or for other reasons. |
+| logs.pod_logs.labels | object | `{"app_kubernetes_io_name":"app.kubernetes.io/name"}` | Loki labels that will be set from the Kubernetes Pod labels. |
```
skl, line 403: This differs from setting static/custom labels in the logs - so I wonder if it's worth stating as part of the value name that this is specifically for k8s label "copying"?

skl (approved):

charts/k8s-monitoring/README.md
petewall: I can make the instructions a bit clearer, too.

PR #627, opened 2024-07-08
Add the ability to set live debugging

petewall:
also clean up some values.yaml sections

### DIFF
--- a/charts/k8s-monitoring/README.md
+++ b/charts/k8s-monitoring/README.md
@@ -150,6 +150,38 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely 
 
 ## Values
 
+### Deployment: [Alloy](https://github.com/grafana/alloy/tree/main/operations/helm/charts/alloy) for Cluster Events Deployment
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| alloy-events.liveDebugging.enabled | bool | `false` | Enable live debugging for the Alloy instance. |
+| alloy-events.logging.format | string | `"logfmt"` | Format to use for writing Alloy log lines. |
+| alloy-events.logging.level | string | `"info"` | Level at which Alloy log lines should be written. |
+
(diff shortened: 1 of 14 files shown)


skl (approved):

stefanandres:
You actually also need to set this in the value.yaml to get it to work, otherwise the validator hook pod will throw the error
```
❯ k logs validate-monitoring-k8s-monitoring
Validating Grafana Alloy config file
Error: /etc/alloy/config.alloy:909:1: block "livedebugging" is at stability level "experimental", which is below the minimum allowed stability level "generally-available". Use --stability.level command-line flag to enable "experimental" features

908 |
909 |   livedebugging {
    |  _^^^^^^^^^^^^^^^
910 | |   enabled = true
911 | | }
    | |_^
912 |   /* Redis */

interrupt received
Error: could not perform the initial load successfully
```

```diff
@@ -203,6 +203,9 @@ alloy:

   serviceMonitor:
     enabled: true
+  alloy:
+    stabilityLevel: experimental
   liveDebugging:
     enabled: true

@@ -263,6 +266,9 @@ alloy-logs:
         maxUnavailable: 100%
   serviceMonitor:
     enabled: true
+  alloy:
+    stabilityLevel: experimental
   liveDebugging:
     enabled: true

@@ -286,6 +292,9 @@ alloy-events:
         memory: 300Mi
   serviceMonitor:
     enabled: true
+  alloy:
+    stabilityLevel: experimental
   liveDebugging:
     enabled: true
``` 

and even then the validator pod fails because it seems you cannot set the --stability.level parameter to the https://github.com/grafana/k8s-monitoring-helm/blob/64e01fcb7460fe6e61f7ad6277ae529a80aaacc2/charts/k8s-monitoring/templates/hooks/validate-configuration.yaml alloy run commands currently. We should add a variable for this.

petewall:
Yeah, thanks for pointing that out. Let me see how we can add that easily.

PR #633, opened 2024-07-11
Make the formatAsJson actually do something and add config validation.

petewall:
Also remove journal extraConfig which also was not used.


charts/k8s-monitoring/templates/alloy_config/_journal_logs_discovery.alloy.txt
```diff
@@ -26,8 +26,9 @@ loki.relabel "journal" {
 }
 
 loki.source.journal "worker" {
-  max_age = {{ .Values.logs.journal.maxAge | default "8h" | quote }}
   path = {{ .Values.logs.journal.path | default "/var/logs/journal" | quote }}
+  format_as_json = {{ .Values.logs.journal.formatAsJson | quote }}
+  max_age = {{ .Values.logs.journal.maxAge | default "8h" | quote }}
```
skl, line 31: Maybe better to specify the default in the values file?

skl (approved):
lgtm

charts/k8s-monitoring/templates/alloy_config/_journal_logs_discovery.alloy.txt
petewall: It is in the values file, but this protects if the value goes empty

PR #608, opened 2024-07-02
allow specifying additional outputs for otlp batch processor

anukul:
motivation: hook into existing otlp receiver and avoid running another one just for exporting data to kafka

e.g.
```yaml
# values.yaml
receivers:
  processors:
    batch:
      output:
        metrics: [otelcol.exporter.kafka.kafka.input]
        logs: [otelcol.exporter.kafka.kafka.input]
        traces: [otelcol.exporter.kafka.kafka.input]
```

PR #640, opened 2024-07-15
feat: allow adding extra outputs

davidspek:
Inspired by https://github.com/grafana/k8s-monitoring-helm/pull/608, this PR adds the ability to specify extra outputs both for the batch processor receiver, as well as at the last stage for metrics and logs after all processing has been done. The use case for this is for a simple way to duplicate metrics, logs and traces across multiple backends. A more complex implementation could be to have the `externalServices.*` values be an array, but that would make the chart much more complicated. Instead, users can use the `extraConfig` value to configure any additional outputs that are then referenced in these new settings.

Example:

```yaml
# values.yaml
receivers:
  processors:
    batch:
      output:
        metrics: [otelcol.exporter.kafka.kafka.input]
        logs: [otelcol.exporter.kafka.kafka.input]
        traces: [otelcol.exporter.kafka.kafka.input]
externalServices:
  prometheus:
    extraOutputs: [otelcol.exporter.otlp.duplicate.input]
  loki:
    extraOutputs: [otelcol.exporter.otlp.duplicate.input]
```

### DIFF
--- a/charts/k8s-monitoring/values.schema.json
+++ b/charts/k8s-monitoring/values.schema.json
@@ -277,6 +277,9 @@
                         "externalLabelsFrom": {
                             "type": "object"
                         },
+                        "extraOutputs": {
+                            "type": "array"
+                        },
                         "host": {
                             "type": "string"
                         },
@@ -392,6 +395,9 @@
                         },
                         "externalLabelsFrom": {
                             "type": "object"
+                        },
+                        "extraOutputs": {
+                            "type": "array"
                         },
                         "host": {
                             "type": "string"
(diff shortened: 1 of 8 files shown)


charts/k8s-monitoring/values.schema.json
```diff
@@ -757,9 +763,6 @@
                         "enabled": {
                             "type": "boolean"
                         },
-                        "extraConfig": {
-                            "type": "string"
-                        },
```
defanator, line 762: What's the reason behind removing this one here?

petewall:
Closing because adding extra outputs (destinations) is now supported in v2.0

PR #646, opened 2024-07-19
Move extraConfigs into the alloy targets

petewall:
Yes, this will eventually be a breaking change that will affect serviceIntegrations. I'm hopeful to have converted those to modules by then.

Create a script to update examples including ensuring that the README has the real values.yaml.


skl (approved):

petewall:
Closing this. This is happening in v2.0

PR #652, opened 2024-07-24
Re-filter after discovery.process to remove any non-kubernetes java processes

petewall:
Before, we were finding java processes by:
1. discovery.kubernetes to get the pods on the same node, and filter by namespace
2. discovery.process to merge those pods with process data
3. discovery.relabel to filter to java processes and set labels

The problem is that `discovery.process` does not just get process info for the pods found in step 1, it returns a list joined with both prods and processes ([docs](https://grafana.com/docs/alloy/latest/reference/components/discovery/discovery.process/#targets-joining)).

This change re-filters in the `discovery.relabel` component to only things that are pods (so, only get kubernetes processes) and the namespace filter again.

### DIFF
--- a/charts/k8s-monitoring/templates/alloy_config/_profiles_java.alloy.txt
+++ b/charts/k8s-monitoring/templates/alloy_config/_profiles_java.alloy.txt
@@ -21,9 +21,14 @@ discovery.process "java_pods" {
 discovery.relabel "java_pods" {
   targets = discovery.process.java_pods.targets
   rule {
-    action = "drop"
-    regex = "Succeeded|Failed|Completed"
     source_labels = ["__meta_kubernetes_pod_phase"]
+    regex = "Succeeded|Failed|Completed"
+    action = "drop"
+  }
+  rule {
+    source_labels = ["__meta_kubernetes_pod_name"]
+    regex = "^$"
+    action = "drop"
   }
   rule {
     source_labels = ["__meta_process_exe"]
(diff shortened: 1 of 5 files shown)


charts/k8s-monitoring/templates/alloy_config/_profiles_java.alloy.txt
```diff
@@ -30,6 +30,18 @@ discovery.relabel "java_pods" {
...
+    regex = ".+"
+    action = "keep"
+  }
+{{- if .Values.profiles.java.namespaces }}
+  rule {
+    source_labels = ["__meta_kubernetes_namespace"]
+    regex = "({{ join "|" .Values.profiles.java.namespaces }})"
+    action = "keep"
+  }
+{{- end }}
```
simonswine, line 44: I don't think it does what you want it to achieve:

This will keep everything that was kept before (java processes) and additionally all pods and additionally the namespace if supplied.

Let me have a think and I propose something what I think will work instead.

simonswine: I am fairly certain the filter for namespaces is already taken care of, but it is nothing I have tested. We might need to deploy to be 100% certain:

The only change I suggest is adding this before the keep role in L28:

```
  rule {
    source_labels = ["__meta_kubernetes_pod_name"]
    regex = "^$"
    action = "drop"
  }
```

petewall: Thanks! Made those changes!

skl (approved):

simonswine (approved):
Thanks for fixing this ❤️ 

LGTM

PR #653, opened 2024-07-24
Add config validation to check for liveDebugging without the experimental stability level

petewall:
We cannot force Alloy to change its stability level, but we can warn the user to enable it.

### DIFF
--- a/charts/k8s-monitoring/README.md
+++ b/charts/k8s-monitoring/README.md
@@ -154,7 +154,7 @@ The Prometheus and Loki services may be hosted on the same cluster, or remotely 
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| alloy-events.liveDebugging.enabled | bool | `false` | Enable live debugging for the Alloy instance. Requires stability level to be set to "public-preview". |
+| alloy-events.liveDebugging.enabled | bool | `false` | Enable live debugging for the Alloy instance. Requires stability level to be set to "experimental". |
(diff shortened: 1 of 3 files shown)


petewall:
```bash
% helm template k8smon charts/k8s-monitoring -f examples/default-values/values.yaml --set alloy.liveDebugging.enabled=true
Error: execution error at (k8s-monitoring/templates/kubernetes-monitoring-telemetry.yaml:2:3): To enable Alloy live debugging, you must set the stabilityLevel to "experimental":
alloy:
  alloy:
    stabilityLevel: experimental

```

skl (approved):

stefanandres:
Hey @petewall - thanks for the implementation, but I think this missed the part where we also need to be able to set the stabilityLevel in the actual helm hook where the configuration is validated by the alloy command, i.e.

https://github.com/grafana/k8s-monitoring-helm/blob/64e01fcb7460fe6e61f7ad6277ae529a80aaacc2/charts/k8s-monitoring/templates/hooks/validate-configuration.yaml#L52-L91

We need to be able to set the stabilityLevel there as well. Without this, the only way to use livedebugging is to set `configValidator.enabled = false`

petewall:
I was able to set the stability level and live debugging and get it working. Can you try again?
This change fixed the validation: https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/templates/hooks/validate-configuration.yaml#L55

stefanandres:
Ah nice, thanks! Tried and worked :)

PR #580, opened 2024-06-19
Use sa token for metrics http autodiscovery and pyroscope pprof scraping

jingkaihe:
This allows alloy to scrape endpoint secured by some rbac termination endpoint such as [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy).

This is especially beneficial to the pyroscope pprof scraping, otherwise we will end up with exposing `pod_ip:6060` cluster wide. with the kubernetes sa token passed in we can instead:

* run the pyroscope endpoint on `127.0.0.1:6060` which is only accessible within the pod itself.
* kube-rbac-proxy authn and authz with token review and subject access review (non resource url `/debug/pprof/*`) and forward the request in the manner of `0.0.0.0:6062 -> 127.0.0.1:6060`
* on the profiler end we can thus grant the sa with non resource url `/debug/pprof/*

end result:

```
root@grafana-k8s-monitoring-alloy-profiles-4lv2x:/# curl "http://11.5.6.27:6062/debug/pprof/"
Unauthorized
root@grafana-k8s-monitoring-alloy-profiles-4lv2x:/# curl -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" "http://11.5.6.27:6062/debug/pprof/"
<html>
<head>
<title>/debug/pprof/</title>
<style>
.profile-name{
        display:inline-block;
```

Please let me know what you think :)

PR #686, opened 2024-08-12
Adding a conditional to enable or disable bearer token in auto discovery

reginaldomoura-jr:
This is to allow users to disable or enable the usage of sa bearer token in auto discovery, this was add on version 1.3.0, but users don't have a option to disable it if this is not needed. Unauthenticated services using Service Account without token or services not behind something like kube-rbac-proxy is returning 400 error.

https://github.com/grafana/k8s-monitoring-helm/pull/580


reginaldomoura-jr:
Hi @petewall and @skl,

Can you help me to figure this issue?

We are not using the Service Account Tokens on our /metrics at this moment, and we are not using a proxy in front of this as well. Can we have a way to disable and enable the token path? I'm doing this proposal, on this MR. Please, let me know if I'm missing something.

Our services is getting 400, when auto discovery try to get the metrics.

petewall (approved):

petewall:
Run `helm-docs` again and I think you'll be set!

reginaldomoura-jr:
Thank You @petewall, I had run `helm-docs`. Tests is passing now.

petewall:
My plan is to merge a few more and drop version 1.5 either today or Monday.

reginaldomoura-jr:
> My plan is to merge a few more and drop version 1.5 either today or Monday.
Thank you! This will help us to update our K8S Monitoring without any impact on metrics being scraped.

PR #691, opened 2024-08-14
Add kube_node_labels by default

petewall:


jewbetcha (approved):
Nice this is great, we've had customers ask for labels in the app

PR #693, opened 2024-08-15
Add support to Beyla Alloy component

marctc:
Added new set of values to configure Alloy to use Beyla components.


petewall (commented):
I like where this is headed.
Does `beyla.ebpf` require special permissions? Special deployment of Alloy?

marctc:
Hey Pete, thanks! `beyla.ebpf` requires `hostPid: true` and `privileged: true`. It can be run with less privileges as well.

PR #696, opened 2024-08-15
Use the release label instead of the app.kubernetes.io/instance label

petewall:
This improves the experience for ArgoCD users.


petewall:
The removal of the instance label check runs the risk of capturing opencost and alloy instances that we weren't before, but maybe that's not terrible.

It should properly capture the instances deployed for ksm, node exporter and windows exporter, though.

jewbetcha (approved):

PR #703, opened 2024-08-23
Add the ability to set headers for all external services.

petewall:
This adds:
* `extraHeaders` and `extraHeadersFrom` fields for the four data destinations that we support and sets the header values for those
    * `extraHeaders` treats values as a string for the Alloy component
    * `extraHeadersFrom` does not, which lets you use other alloy components, like environment variables, or configmap references


snowzach:
Holy crap.. that was fast... Thanks!!!!!

petewall:
> Holy crap.. that was fast... Thanks!!!!!

You're too kind! Let me reviews from my colleagues and I'll get this merged and released.

jewbetcha (approved):

PR #711, opened 2024-09-02
Allow default metrics used by kubernetes-mixin for alert rules

stefanandres:
By default allow metrics that are used by the common kubernetes-mixin. These metrics are needed for the [KubePersistentVolumeFillingUp alert rule](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/8c0479c02d05f479fafdcce56ad09ea7a264998a/alerts/storage_alerts.libsonnet#L20-L32)


skl (approved):
👍

stefanandres:
Tests should be fixed now

PR #722, opened 2024-09-07
feat: prometheus sigv4 support

### LINKED ISSUE #664: Allow export metrics to Amazon Managed Prometheus
Hello,
as far as I can see the chart doesn't allow to export metrics to a managed Prometheus service like Amazon Managed Prometheus.

I'd like this feature to be added.

Relevant configuration:
https://github.com/grafana/k8s-monitoring-helm/blob/7b9fc9cddb58c099955c1c62488d0eb1e0b54dd1/charts/k8s-monitoring/templates/alloy_config/_metrics_service_remote_write.alloy.txt#L6

The possibility to specify a different auth method is also missing:
https://github.com/grafana/k8s-monitoring-helm/blob/7b9fc9cddb58c099955c1c62488d0eb1e0b54dd1/charts/k8s-monitoring/templates/alloy_config/_metrics_service_remote_write.alloy.txt#L11-L13

a-abella:
Fixes #664 

**References:**

* [prometheus.remote_write](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#sigv4-block)

**Changes:**

Prometheus remote write SigV4 support.

**Notes:**

* Disables setting of `X-Scope-OrgId` header when `authMode: sigv4`, as AWS Managed Prometheus (the only Prometheus implementation behind SigV4, to my knowledge) does not support this header and denies requests when set.
* Follows the Alloy `prometheus.remote_write` convention where no field of the `sigv4` block is Required, however it is worth noting (and is noted in the helm-docs) that `region` is a required field if it cannot be inferred from the default credentials chain. [otelcol.auth.sigv4](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.auth.sigv4/#arguments) handles this more intelligently and tries to infer the region from the endpoint URL if no region is passed and no default credentials chain is present.
* This is a resubmission of #720, noticed something was wrong and wanted a clean PR, apologies if it generated any excess noise.


petewall (approved):

a-abella:
Rebased and re-tested on latest 1.5.4

PR #728, opened 2024-09-09
Adding AWS SigV4 auth option for Prometheus

pydolan:
This adds a "sigv4" auth option for remote-write, which I need to send data to an AWS Managed Prometheus instance. I mirror the sigv4 options listed on Alloy's docs: https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#sigv4-block

(FWIW, I had to add `--validate-maintainers=false` to `ct` for it to not error when run locally)


pydolan:
Just realized that this is a duplicate of #722, so I will close this in favor of it

PR #739, opened 2024-09-18
Allow for the pod logs annotation to actually be configurable

petewall:


jewbetcha (approved):
No examples generated for this change?

petewall:
> No examples generated for this change?

Examples were regenerated, just no changes to the output.

PR #740, opened 2024-09-18
The presence of __tmp_port was breaking deduplication rules.

petewall:
When annotation autodiscovery was written, it was understood that labels starting with `__` wouldn't affect the deduplication rules, and just labels and `__address__` was important. However, all labels, except `__meta` labels are used in deduplicating targets, even though `__` labels are dropped before scraping.


mattdurham (approved):
LGTM

mattdurham:
Additional context: There are special modifiers like `__scrape_interval` that create separate targets even if labels and address are the same. So __ are honored has a catchall, changing to only take into account the special ones is likely a good change but also a breaking one for Alloy/prometheus.

PR #788, opened 2024-10-11
fix: prom destination template config error and change externalLabels to extraLabels

fplantinga-guida:
Noticed that the examples showed extraLabels instead of externalLabels. When I changed my values to externalLabels I was getting an EOF config error when adding the labels. The block was not closed correctly. 

I'm also happy to change the documentation to externalLabels as the argument is called external_labels in the documenation. https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#arguments

### DIFF
--- a/charts/k8s-monitoring/docs/examples/extra-rules/alloy-metrics.alloy
+++ b/charts/k8s-monitoring/docs/examples/extra-rules/alloy-metrics.alloy
@@ -41,6 +41,10 @@ prometheus.remote_write "prometheus" {
       action = "drop"
     }
   }
+  external_labels = {
+    site = "lab2",
+    region = env("REGION"),
+  }
 }
 // Feature: Cluster Metrics
 declare "cluster_metrics" {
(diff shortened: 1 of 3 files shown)


petewall (approved):
Thanks for catching that!

petewall:
Run "make build" and it should pick up the changes into the examples

fplantinga-guida:
Make build didn't work for me. I modified the examples manually

petewall:
> I'm also happy to change the documentation to externalLabels as the argument is called external_labels in the documenation. [grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write#arguments](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#arguments)

Labeling them "extraLabels" is unique to this Helm chart, so nothing outside of this repo should get a change.

The use of "externalLabels" in Alloy is an idiom that comes from Prometheus. In Prometheus, it was called external labels, because those were the labels that you added when sending metrics to some location external to itself. This chart does not deploy prometheus, nor is it a database. All metrics are delivered "externally", so I thought it would be friendlier to newcomers to call them "extra" labels instead.

PR #799, opened 2024-10-15
Set apiserver enabled default to false in feature-cluster-metrics values

mtweten:
I noticed this when playing around with the k8s-monitoring v2 chart - the apiserver enabled field wasn't set to default to `false` in the values yaml, so the generated schema only allowed "null" for the value.

### DIFF
--- a/charts/feature-cluster-metrics/values.yaml
+++ b/charts/feature-cluster-metrics/values.yaml
@@ -187,10 +187,10 @@ cadvisor:
 
 # API Server metrics gather information about the Kubernetes API Server.
 apiServer:
-  # -- Scrape metrics from the API Server
-  # @default -- false
+  # -- (bool) Scrape metrics from the API Server.
+  # @default -- `false`
   # @section -- API Server
-  enabled: false
+  enabled: 
 
   # -- How frequently to scrape metrics from the API Server
   # Overrides metrics.scrapeInterval
--- a/charts/feature-cluster-metrics/values.yaml
+++ b/charts/feature-cluster-metrics/values.yaml
@@ -190,7 +190,7 @@ apiServer:
   # -- Scrape metrics from the API Server
   # @default -- false
   # @section -- API Server
-  enabled:
+  enabled: false
 
   # -- How frequently to scrape metrics from the API Server
   # Overrides metrics.scrapeInterval
--- a/charts/feature-cluster-metrics/values.yaml
+++ b/charts/feature-cluster-metrics/values.yaml
@@ -190,7 +190,7 @@ apiServer:
   # -- (bool) Scrape metrics from the API Server.
   # @default -- `false`
   # @section -- API Server
-  enabled: 
+  enabled:
 
   # -- How frequently to scrape metrics from the API Server
   # Overrides metrics.scrapeInterval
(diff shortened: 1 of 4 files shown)


petewall:
There's actually a very good reason this is set to empty, rather than false.
Because I want to be able to use both:
```
controlPlane: true/false
apiServer.enabled: true/false
```

To set individual features.

| `controlPlane` | `apiServer.Enabled` | API Server enabled? |
|----------------|---------------------|-------|
| `false`               | `null`                         | No |
| `false`               | `false` | No | 
| `false`               | `true` | Yes (and only API server) | 
| `true` | `null` | Yes (along with the other control plane options) |
| `true` | `false` | No (but the other control plane options are) |
| `true` | `true` | Yes |

So, you can see, I want to differentiate between the default and the user explicitly setting "false".

If you had a config with these values:
```yaml
controlPlane: true
apiServer:
  enabled: false
```

I'd want it to do what's expected.

petewall:
All that being said, I think there are improvements to be made to the docs here. I'll try and put them into the comments.

charts/feature-cluster-metrics/values.yaml
```diff
@@ -190,7 +190,7 @@ apiServer:
   # -- Scrape metrics from the API Server
   # @default -- false
   # @section -- API Server
-  enabled:
+  enabled: false
```
petewall, line 193:
```suggestion
  enabled: 
```

charts/feature-cluster-metrics/values.yaml
```diff
@@ -190,7 +190,7 @@ apiServer:
   # -- Scrape metrics from the API Server
```
petewall, line 190:
```suggestion
  # -- (bool) Scrape metrics from the API Server.
```

charts/feature-cluster-metrics/values.yaml
```diff
@@ -190,7 +190,7 @@ apiServer:
...
   # @default -- false
```
petewall, line 191:
```suggestion
  # @default -- `false`
```

petewall (changes requested):

petewall:
Also modify `charts/feature-cluster-metrics/schema-mods/types-and-enums.json` to allow the boolean value:
```json
```

mtweten:
Ah yep, I get your use case. I had just gotten the validation error when trying to set apiServer.enabled to true since the schema json was only allowing null. I just didn't know the proper way to fix it besides the default to boolean value - didn't know about the schema-mods!

petewall:
> Ah yep, I get your use case. I had just gotten the validation error when trying to set apiServer.enabled to true since the schema json was only allowing null. I just didn't know the proper way to fix it besides the default to boolean value - didn't know about the schema-mods!

`schema-mods` is my way of making changes since the tool that generates the schema from the values.yaml file isn't very smart, but it's the best I've found so far.

Chances are there are more examples of this behavior in the other control plan components (scheuler, proxy, etc...), if you're up for it, I'd appreciate help fixing those, too! Thanks again for looking into this chart!

mtweten:
> > Ah yep, I get your use case. I had just gotten the validation error when trying to set apiServer.enabled to true since the schema json was only allowing null. I just didn't know the proper way to fix it besides the default to boolean value - didn't know about the schema-mods!
> 
> `schema-mods` is my way of making changes since the tool that generates the schema from the values.yaml file isn't very smart, but it's the best I've found so far.
> 
> Chances are there are more examples of this behavior in the other control plan components (scheuler, proxy, etc...), if you're up for it, I'd appreciate help fixing those, too! Thanks again for looking into this chart!

Yep, I'm going to go ahead and update the control plane components with the same sets of changes we're applying to apiServer.

petewall (approved):

PR #805, opened 2024-10-16
Add the ability to get config remotely

petewall:
Notes to reviewers:
The only files I'd like you to review are:
* [charts/k8s-monitoring/values.yaml](https://github.com/grafana/k8s-monitoring-helm/pull/805/files#diff-cd22d083c6f49784d198f55b990feb9ae245abf1353dd56a379ec880126f4488)
* [charts/k8s-monitoring/templates/collectors/_collector_remoteConfig.tpl](https://github.com/grafana/k8s-monitoring-helm/pull/805/files#diff-eb197ffcf7acf0146cf41ef2f18bcc2431c77b3c92115853468b38479be359b0)
* the files inside of charts/k8s-monitoring/docs/examples/remote-config

So much of the rest are just:
* Refactoring secret handling to work for things other than destinations
* Adding the concept of "platform tests" which are special integration tests that utilize external dependencies like Grafana Cloud.


charts/k8s-monitoring/values.yaml
```diff
@@ -214,6 +214,68 @@ alloy-metrics:
...
+
+      # -- The username to use for the remote config server.
+      # @section -- Collectors - Alloy Metrics
+      username: ""
+      # -- The key for storing the username in the secret.
+      # @section -- Collectors - Alloy Metrics
+      usernameKey: "username"
+      # -- Raw config for accessing the password.
+      # @section -- Collectors - Alloy Metrics
+      usernameFrom: ""
```
skl, line 240: Do these get used somewhere?

skl (approved):

charts/k8s-monitoring/values.yaml
petewall: Yeah, because this line:
```
username = {{ include "secrets.read" (dict "object" . "key" "auth.username" "nonsensitive" true) }}
```
Actually does a ton.

If your values are:
```
auth: { username = "myuser" }
```
Then, this will create a K8s secret, with `username: myuser` and then the alloy config references that secret.

If your values are:
```
auth: { username = "myuser" }
secret: { embed = true }
```
Then the string "myuser" is set in the alloy config and no secret is created.

Finally, if your values are:
```
auth: { usernameFrom = env("MY_USERAME") }
```
Then it injects the raw string env("MY_USERAME") as the value for the username.

All of this is handled by the template functions.

petewall: That's not even mentioning these values:
```
auth: { usernameKey = "user" }
secret: { create = false, name = mysecret }
```
Which then references another pre-created secret with a custom key in that secret.

PR #817, opened 2024-10-25
Add MySQL integration

petewall:
Log handling requires the pod logs feature. It adds a relabeling rule during discovery and a processing stage block.


petewall:
Cert manager is broken until https://github.com/grafana/alloy-modules/pull/40 is merged.

PR #832, opened 2024-10-31
Fix typo, improve destination validations

petewall:


charts/k8s-monitoring/schema-mods/definitions/invalid-destination.schema.json
petewall, line 1: This allows for invalid types to pass schema validation... only to be caught by the template validation. It greatly improves the quality of the error message if you're missing a type or use an invalid type.

charts/k8s-monitoring/templates/validations.yaml
petewall, line 1: Doing the destination validation before feature validation to catch these issues earlier.
For example, if your values file had this:
```
destinations:
  - name: prom
    type: prom   # invalid type
clusterMetrics:
  enabled: true
```

The error message will say that Cluster Metrics requires a metrics destination, not that "prom" is not a valid destination type.

skl (approved):

PR #843, opened 2024-11-06
Create the option for loading alloy modules embedded in the chart, rather than from git

petewall:
Signed-off-by: Pete Wall <pete.wall@grafana.com>


charts/feature-cluster-metrics/templates/_helpers_modules.tpl
```diff
@@ -0,0 +1,24 @@
...
+{{- $pathParts := regexSplit "/" .path -1 }}
+{{- $configMapName := printf "%s-alloy-module-%s" $.Release.Name (index $pathParts 1) }}
+{{- $moduleFile := (slice $pathParts 2) | join "_" }}
+remote.kubernetes.configmap {{ .name | quote }} {
+  name = {{ $configMapName | quote }}
+  namespace = {{ .Release.Namespace | quote }}
+}
+
+import.string {{ .name | quote }} {
+  content = remote.kubernetes.configmap.{{ .name }}.data[{{ $moduleFile | quote }}]
```
skl, line 14: `$moduleFile` is presumably a path visible to the Alloy pod used to load the module from disk?

skl (approved):
Nice.

The modules you've chosen to vendor - is that based on the modules which are used/available by default in the chart? Presumably this works toward having an "air-gapped"/self-contained build of the chart.

Is it "all or nothing" for the git/configmap choice? i.e. if I choose `configMap` then I need to ensure that all of the chart's modules are defined in configmaps and all of their files are available on disk, before I start to add more modules? Does the vendoring help with that?

Thinking aloud a bit here but love the direction this is going 😄

charts/feature-cluster-metrics/templates/_helpers_modules.tpl
petewall: Yep. `remote.kubernetes.configmap` creates an array of string objects, referenced by their keys.

petewall:
> is that based on the modules which are used/available by default in the chart?

It is, but not via automation. It's picked by the vendir.yml file. It's manual for now, but it could be extracted from the code at some point.

> Is it "all or nothing" for the git/configmap choice? i.e. if I choose configMap then I need to ensure that all of the chart's modules are defined in configmaps and all of their files are available on disk, before I start to add more modules? Does the vendoring help with that?

Currently, yes, it is all or nothing. It's a first drop. At some point, making the module location configurable for each would be nice, especially if we need to override for whatever reason. But this satisfies one community tester's problem of "my cluster can't access git".

PR #847, opened 2024-11-07
add missing tls config for loki writer

ohayak:
add missing tls config for loki writer


charts/k8s-monitoring/templates/destinations/_destination_loki.tpl
```diff
@@ -33,6 +33,21 @@ loki.write {{ include "helper.alloy_name" .name | quote }} {
...
     bearer_token = {{ include "secrets.read" (dict "object" . "key" "auth.bearerToken") }}
 {{- end }}
+
+{{- if .tls }}
+    tls_config {
+      insecure_skip_verify = {{ .tls.insecureSkipVerify | default false }}
+      {{- if eq (include "secrets.usesSecret" (dict "object" . "key" "tls.ca")) "true" }}
+      ca_pem = {{ include "secrets.read" (dict "object" . "key" "tls.ca" "nonsensitive" true) }}
+      {{- end }}
+      {{- if eq (include "secrets.usesKubernetesSecret" .) "true" }}
```
petewall, line 43:
```suggestion
      {{- if eq (include "secrets.usesSecret" (dict "object" . "key" "tls.cert")) "true" }}
```

charts/k8s-monitoring/templates/destinations/_destination_loki.tpl
```diff
@@ -33,6 +33,21 @@ loki.write {{ include "helper.alloy_name" .name | quote }} {
...
+      cert_pem = {{ include "secrets.read" (dict "object" . "key" "tls.cert" "nonsensitive" true) }}
+      {{- end }}
+      {{- if eq (include "secrets.usesKubernetesSecret" .) "true" }}
```
petewall, line 46:
```suggestion
      {{- if eq (include "secrets.usesSecret" (dict "object" . "key" "tls.key")) "true" }}
```

petewall (changes requested):
This is looking good. Thanks for adding it.
run `make clean build test` at the top level and it should also include any generated file changes.

Note that `usesKubernetesSecret` was a bug in other files. It should be `usesSecret`.

PR #893, opened 2024-11-12
feat(loki): add tls config

ohayak:
added tls config for loki


petewall:
Looks like the `make clean` ran without a `make build`. Can you try again?

petewall:
Fixed in #911

PR #940, opened 2024-11-22
Fix the migration guide's extaConfg table

### LINKED ISSUE #939: v2: Migration.md extraConfig erroneous v2 path
When migrating from v1 to v2 the Migration guide states that the extraConfig keys is now scoped under the alloy-XXXX keys but the table at the beginning of the paragraph seems a copy/paste of the integration one, ie the previous paragraph.

In short: "Steps to take" is correct but the table is not.

Reference: https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/docs/Migration.md#extra-configs

petewall:
Fixes #939


petewall:
Thanks for catching this, @marcomusso

PR #948, opened 2024-11-26
Add write-ahead log settings for prometheus destinations

petewall:


charts/k8s-monitoring/templates/destinations/_destination_prometheus.tpl
```diff
@@ -101,7 +101,7 @@ prometheus.remote_write {{ include "helper.alloy_name" .name | quote }} {
     }
 {{- end }}
     send_native_histograms = {{ .sendNativeHistograms | default false }}
-{{- if .queueConfig }}
```
petewall, line 104: Removed because this is always true

skl (approved):
lgtm. Does this relate implicityly/automatically in any way to the creation of a PV+PVC for the WAL? Or does that need additional config?

petewall:
> lgtm. Does this relate implicityly/automatically in any way to the creation of a PV+PVC for the WAL? Or does that need additional config?

Unfortunately no, it just controls the WAL settings for alloy. Without making any changes to the PV/PVC, it'll just write to the container filesystem

PR #953, opened 2024-11-26
Add Node Logs feature, which currently only handles journal logs

petewall:


charts/feature-node-logs/templates/_module.alloy.tpl
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  loki.source.journal "worker" {
+    path = {{ .Values.journal.path | quote }}
+    format_as_json = {{ .Values.journal.formatAsJson }}
+    max_age = {{ .Values.journal.maxAge | quote }}
+    relabel_rules = loki.relabel.journal.rules
+    labels = {
+      job = {{ .Values.journal.jobLabel | quote }},
+      instance = env("HOSTNAME"),
```
skl, line 35: Do you know if the hostname here would generally match the k8s node name?

skl (approved):

charts/feature-node-logs/templates/_module.alloy.tpl
petewall: I believe it matches the Node name where the Alloy pod lives.

PR #979, opened 2024-12-09
Fix errors in clustermetrics validation

eliihen:
This PR fixes the following error:

When using `urlFrom` in the prometheus destination, it only uses the `url` field and throws due to url being `Nil`:

```
$ helm dep up; helm template observability .
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "grafana" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Deleting outdated charts
Error: template: observability/charts/k8s-monitoring/charts/k8s-monitoring/templates/validations.yaml:9:6: executing "observability/charts/k8s-monitoring/charts/k8s-monitoring/templates/validations.yaml" at <include (printf "features.%s.validate" $feature) $>: error calling include: template: observability/charts/k8s-monitoring/charts/k8s-monitoring/templates/features/_feature_cluster_metrics.tpl:71:43: executing "features.clusterMetrics.validate" at <$destinationUrl>: invalid value; expected string

```

### DIFF
--- a/charts/k8s-monitoring/templates/features/_feature_cluster_metrics.tpl
+++ b/charts/k8s-monitoring/templates/features/_feature_cluster_metrics.tpl
@@ -66,7 +66,7 @@ cluster_metrics "feature" {
       {{- if eq $destinationName $.Values.clusterMetrics.opencost.metricsSource }}
         {{- $destinationFound = true }}
         {{- $destination := index $.Values.destinations $index }}
-        {{- $destinationUrl := $destination.url }}
+        {{- $destinationUrl := $destination.url | default $destination.fromUrl | default "" }}
         {{- $openCostMetricsUrl := $destinationUrl }}
         {{- if regexMatch "/api/prom/push" $destinationUrl }}
           {{- $openCostMetricsUrl = (regexReplaceAll "^(.*)/api/prom/push$" $destinationUrl "${1}/api/prom") }}


petewall:
Thanks for catching this. There's a real bug here. I think we'll need to think a bit more about this.

In the URL case:
```yaml
- type: prometheus
  url: http://prom.example.com/api/prom
```

the OpenCost validation tries to build a url for OpenCost to send promql queries to.

In the URL From case:
```yaml
- type: prometheus
  urlFrom: sys.env("PROM_URL")
```

It can't turn that into a recommendation. Because of the OpenCost Helm chart, I can't use the same urlFrom semantic. What needs to happen is an "escape hatch" part to the validation that just says:

You wanna configure OpenCost? This is what you'll need:
```yaml
clusterMetrics:
  opencost:
    opencost:
      prometheus:
        username: <basic auth username>
        password: <basic auth password>
        external:
          url: <Prometheus query URL>
```

petewall:
Can you retest with the latest rc? It should handle things better now.

petewall:
Closing. A newer version should make this work now.

eliihen:
Apologies for the delay, I've been offline over the christmas holidays. I testet with the 2.0 release and it works now 🙂

petewall:
Wonderful! Thanks for letting me know!

PR #980, opened 2024-12-09
Add example and platform test to check the otlp gateway

petewall:
* Adds an example on how to use the [Grafana Cloud OTLP gateway](https://grafana.com/docs/grafana-cloud/send-data/otlp/send-data-otlp/) (also called the otlp endpoint).
* Adds a platform test that sends metrics and logs to Grafana Cloud, using the gateway.
* Adds a `valueFrom` field to the OTLP destination processors, that lets be set values based on environment variables.

### DIFF
--- a/charts/k8s-monitoring/docs/examples/destinations/otlp-gateway/values.yaml
+++ b/charts/k8s-monitoring/docs/examples/destinations/otlp-gateway/values.yaml
@@ -1,0 +1,28 @@
+---
+cluster:
+  name: otlp-gateway-test
+
+destinations:
+  - name: otlp-gateway
+    type: otlp
+    url: https://otlp-gateway-my-region.grafana.net/otlp
+    protocol: http
+    auth:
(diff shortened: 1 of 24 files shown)


charts/k8s-monitoring/docs/examples/destinations/otlp-gateway/values.yaml
```diff
@@ -0,0 +1,28 @@
...
+    type: otlp
+    url: https://otlp-gateway-my-region.grafana.net/otlp
+    protocol: http
+    auth:
+      type: basic
+      username: my-gateway-username
+      password: my-gateway-password
+    metrics: {enabled: true}
+    logs: {enabled: true}
+    traces: {enabled: true}
```
jewbetcha, line 16: Just a question for my own understanding, is the `{enabled: true}` syntax specific to OTLP?

jewbetcha (approved):

PR #988, opened 2024-12-10
de-duplicate targets from portNumber annotation

j6nca:
We ran into an issue where sidecar containers (i.e vault agent) get registered as targets due to inheriting the pod's annotations for autodiscovery and resulting in alloy scraping the pod multiple times. This fix does a similar check as the portName check above it, and removes these additional targets if the `portNumber` annotation is specified and the container does not have a corresponding port number exposed.

### DIFF
--- a/charts/feature-annotation-autodiscovery/templates/_module.alloy.tpl
+++ b/charts/feature-annotation-autodiscovery/templates/_module.alloy.tpl
@@ -49,6 +49,18 @@ declare "annotation_autodiscovery" {
       target_label = "__tmp_port"
     }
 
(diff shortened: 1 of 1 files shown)


petewall:
Please try things again with the latest (2.0.0-rc.13). I fixed an issue with duplicate port scraping.
I actually don't want the portNumber annotation to check for exposed ports, because sometimes, the metrics port is not exposed, but we need to explicitly set it to something. Only checking for exposed ports prevents this option.

petewall:
I'm going to close this, as we can't take a solution that requires the port number to be exposed.
Try with the latest version and see if that helps!

PR #1012, opened 2024-12-18
Improved validators for OpenCost prometheus URLs and basic auth

### LINKED ISSUE #1005: Add check if `basic` auth is specified, but no username or password is set
If a destination is accidentally defined like:
```
- name: destination
  type: anything
  auth:
    type: basic:
    basic:
      username: myuser
      password: mypassword
```

It *looks* right, but `username` and `password` aren't inside of a `basic` section. We could add a check for `auth.type = basic` without username or password defined and warn the user (with keeping aware of things like external secrets).

petewall:
Fixes #1005


skl (approved):
💯

PR #1019, opened 2024-12-19
Fix building of the allow list.

petewall:
This was breaking when the yaml allow list file contained '---', which was treated as separate lists and not joined together.


rlankfo (approved):
LGTM 🏄‍♂️

PR #1026, opened 2024-12-20
Added Grafana Integration

bentonam:


charts/k8s-monitoring/charts/feature-integrations/default-allow-lists/grafana.yaml
petewall, line 1: If this is empty (meaning, we keep all metrics), we should remove it.

charts/k8s-monitoring/charts/feature-integrations/integrations/grafana-values.yaml
```diff
@@ -0,0 +1,84 @@
+---
+# -- Name for this Loki instance.
```
petewall, line 2:
```suggestion
# -- Name for this Grafana instance.
```

petewall: Lots of other places in this file, too.

charts/k8s-monitoring/charts/feature-integrations/integrations/grafana-values.yaml
```diff
@@ -0,0 +1,84 @@
...
+  # Overrides global.maxCacheSize
+  # @default -- `100000`
+  # @section -- Metric Processing Settings
+  maxCacheSize:
+
+  # Adjustments to the scraped metrics to filter the amount of data sent to storage.
+  tuning:
+    # -- Filter the list of metrics from Grafana Loki to the minimal set required for the Grafana Loki integration.
+    # @section -- Metric Processing Settings
+    useDefaultAllowList: true
```
petewall, line 47: Just like the empty file, if we don't have an allow list, then this should be removed.

charts/k8s-monitoring/charts/feature-integrations/default-allow-lists/grafana.yaml
bentonam: Removed

charts/k8s-monitoring/charts/feature-integrations/integrations/grafana-values.yaml
bentonam: Fixed

charts/k8s-monitoring/charts/feature-integrations/integrations/grafana-values.yaml
bentonam: Removed

petewall (approved):

PR #1045, opened 2025-01-02
Update KSM, Node Exporter, and OpenCost

petewall:


skl (approved):
not sure why this'd cause the test to fail, looks like something might not have a deterministic ordering somewhere?

petewall:
> not sure why this'd cause the test to fail, looks like something might not have a deterministic ordering somewhere?

That's absolutely the cause... piggy-backing a fix for that into here.

PR #1046, opened 2025-01-02
Move the feature charts as directories inside of the k8s-monitoring chart

petewall:
This should improve chart dependency handling


petewall:
We can safely ignore the failures for `feature-pod-logs` and `feature-profiling` because those charts don't exist in this change.

PR #1055, opened 2025-01-06
Make a regular action to check for alloy modules updates

petewall:


bentonam (approved):
LGTM

Only additional feedback would be running this workflow as a dispatch/cron is fine, but it might be worthwhile to also update the grafana/alloy-modules repo to explicitly call this workflow whenever a new tag is published.  That way this workflow doesn't need to run on a cron at all

petewall:
> LGTM
> 
> Only additional feedback would be running this workflow as a dispatch/cron is fine, but it might be worthwhile to also update the grafana/alloy-modules repo to explicitly call this workflow whenever a new tag is published. That way this workflow doesn't need to run on a cron at all

Eventually, yeah. That's a good idea.

PR #1057, opened 2025-01-07
Add test showing how to use extraConfig and statsd

petewall:
Originally made to test a customer situation, but found it useful to show an alternative use of the alloy-receiver.

### DIFF
--- a/charts/k8s-monitoring/tests/integration/statsd/deployments/query-test.yaml
+++ b/charts/k8s-monitoring/tests/integration/statsd/deployments/query-test.yaml
@@ -1,0 +1,42 @@
+---
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: k8s-monitoring-test
+spec:
+  interval: 1m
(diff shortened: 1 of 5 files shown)


charts/k8s-monitoring/tests/integration/statsd/deployments/query-test.yaml
```diff
@@ -0,0 +1,42 @@
+---
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: k8s-monitoring-test
+spec:
+  interval: 1m
+  url: https://github.com/grafana/k8s-monitoring-helm
+  ref:
+    branch: hackathon-2024-12-cluster-o-nimbus
```
skl, line 10: is this the right branch ref?

skl (approved):
🚀

charts/k8s-monitoring/tests/integration/statsd/deployments/query-test.yaml
petewall: Haha, yep, gotta change that to main.

PR #1060, opened 2025-01-07
Add the Node Logs feature to a test.

petewall:
Also add a run-test target to platform tests Makefiles to make it easier to run locally.

Adding them to the EKS test because the GKE Autopilot cluster or any Kind cluster wont have journal logs.

### DIFF
--- a/charts/k8s-monitoring/tests/platform/eks-with-windows/.rendered/output.yaml
+++ b/charts/k8s-monitoring/tests/platform/eks-with-windows/.rendered/output.yaml
@@ -937,6 +937,7 @@ data:
     # TYPE grafana_kubernetes_monitoring_feature_info gauge
     grafana_kubernetes_monitoring_feature_info{deployments="kube-state-metrics,node-exporter,windows-exporter", feature="clusterMetrics", sources="kubelet,kubeletResource,cadvisor,kube-state-metrics,node-exporter,windows-exporter", version="1.0.0"} 1
     grafana_kubernetes_monitoring_feature_info{feature="clusterEvents", version="1.0.0"} 1
+    grafana_kubernetes_monitoring_feature_info{feature="nodeLogs", version="1.0.0"} 1
     grafana_kubernetes_monitoring_feature_info{feature="podLogs", method="volumes", version="1.0.0"} 1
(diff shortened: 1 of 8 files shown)


charts/k8s-monitoring/tests/platform/eks-with-windows/deployments/query-test.yaml
```diff
@@ -83,6 +85,10 @@ spec:
           - query: count_over_time({cluster="$CLUSTER", job="integrations/kubernetes/eventhandler"}[1h])
             type: logql
 
+          # Node logs
+          - query: count_over_time({cluster="$CLUSTER", job="integrations/kubernetes/journal"}[1h])
+            type: logql
+
           # Pod logs
-          - query: count_over_time({cluster="$CLUSTER", job!="integrations/kubernetes/eventhandler"}[1h])
+          - query: count_over_time({cluster="$CLUSTER", job="default/kube-state-metrics"}[1h])
```
skl, line 93: Is it less fragile to exclude both events and node logs?

Mainly thinking about this query from the k8s app perspective and I know it doesn't have much impact here... unless you have a test with a different KSM job label or KSM logging turned off/filtered.

Fine either way, just made me think! 😄 

```suggestion
          - query: count_over_time({cluster="$CLUSTER", job!~"integrations/kubernetes/eventhandler|integrations/kubernetes/journal"}[1h])
```

charts/k8s-monitoring/tests/platform/eks-with-windows/.rendered/output.yaml
```diff
@@ -979,6 +980,159 @@ data:
       namespace = "default"
     }
     
+    // Feature: Node Logs
```
skl, line 983: Presume you'd need to maintain this every time the node logs module is updated?

skl (approved):

charts/k8s-monitoring/tests/platform/eks-with-windows/.rendered/output.yaml
petewall: Alloy modules are synced and kept locally. So whenever they're updated, it'll update the examples and rendered test deployments. No manual steps required.

PR #1095, opened 2025-01-10
Fix otlp-gateway readme typo

fionaliao:
Teeny fix - `thte` -> `the`


petewall (approved):

petewall:
OOh thanks!
Can you also update the `description.txt` file in the same directory? That's the source of that content lives, actually!

PR #1118, opened 2025-01-14
Add kube_configmap_info

petewall:
Also extract kube-state-metrics' allow list to a common file path


petewall:
cc @gouthamve

gouthamve:
Thank you <3!

PR #1120, opened 2025-01-15
Discovery improvements

petewall:
This change uses a new version of alloy modules that improves discovery for kube-state-metrics, node-exporter, and windows-exporter. This make it more efficient on most clusters and it will also make it functional on openshift clusters.


charts/k8s-monitoring/charts/feature-cluster-metrics/README.md
```diff
@@ -197,7 +198,10 @@ Be sure perform actual integration testing in a live environment in the main [k8
 | kube-state-metrics.metricsTuning.excludeMetrics | list | `[]` | Metrics to drop. Can use regular expressions. |
 | kube-state-metrics.metricsTuning.includeMetrics | list | `[]` | Metrics to keep. Can use regular expressions. |
 | kube-state-metrics.metricsTuning.useDefaultAllowList | bool | `true` | Filter the list of metrics from Kube State Metrics to a useful, minimal set. |
+| kube-state-metrics.namespace | string | `""` | Namespace to locate Node Exporter pods. If `deploy` is set to `true`, this will automatically be set to the namespace where this Helm chart is deployed. |
```
skl, line 201:
```suggestion
| kube-state-metrics.namespace | string | `""` | Namespace to locate kube-state-metrics pods. If `deploy` is set to `true`, this will automatically be set to the namespace where this Helm chart is deployed. |
```

skl (approved):

PR #1130, opened 2025-01-16
Feat/integration updates

bentonam:


charts/k8s-monitoring/charts/feature-integrations/templates/_integration_grafana_logs.tpl
```diff
@@ -19,28 +19,28 @@
...
   replacement = "integrations/grafana"
 }
 rule {
-  source_labels = {{ $labelList | sortAlpha | toJson }}
+  source_labels = {{ $labelList | toJson }}
   separator = ";"
-  regex = {{ $valueList | sortAlpha | join ";" | quote }}
+  regex = {{ $valueList | join ";" | quote }}
   target_label = "instance"
   replacement = {{ $instance.name | quote }}
```
petewall, line 45: We gotta do something about predictable layout for this. We had the `sortAlpha` in there before because it wasn't deterministic and it failed autotests pretty regularly.

Maybe instead of `range $k, $v := .labelSelectors` we do something more like `range $key := (keys .labelSelectors | sort)`?

petewall (approved):

PR #1132, opened 2025-01-16
feat: updated node exporter metrics integration list

bentonam:
The integration allow list was missing several metrics that are used by the cloud integration, this corrects that.  

I downloaded all of the dashboards and rules in the integration and used `mimirtool analyze dashboard` and `mimirtool `analyze rule-file` to extract all of the metrics that were used and make a unique list


petewall (approved):
🚀

PR #1150, opened 2025-01-21
Update the migration guide to discuss how to rename data destination …

petewall:
…names with extraConfig section.

### DIFF
--- a/charts/k8s-monitoring/docs/Migration.md
+++ b/charts/k8s-monitoring/docs/Migration.md
@@ -298,7 +298,11 @@ If using Prometheus Operator objects, `metrics.podMonitors.enabled`, `metrics.pr
 ### Integrations
 
 Integrations are a new feature in v2.0 that allow you to enable and configure additional data sources, but this also
-includes the Alloy metrics that were previously part of `v1`.
+includes the Alloy metrics that were previously part of `v1`. Some service integrations were previously defined in the
+`extraConfig` and `logs.extraConfig` sections. If you are using those fields for any of the new build-in integrations,
+you can replace your `extraConfig` to the new `integrations` feature.
+
+#### Built-in integrations
 
 | Integration  | v1.x setting                       | v2.0 setting                | Notes |
 |--------------|------------------------------------|-----------------------------|-------|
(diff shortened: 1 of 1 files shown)


charts/k8s-monitoring/docs/Migration.md
```diff
@@ -298,7 +298,11 @@ If using Prometheus Operator objects, `metrics.podMonitors.enabled`, `metrics.pr
 ### Integrations
 
 Integrations are a new feature in v2.0 that allow you to enable and configure additional data sources, but this also
-includes the Alloy metrics that were previously part of `v1`.
+includes the Alloy metrics that were previously part of `v1`. Some service integrations were previously defined in the
+`extraConfig` and `logs.extraConfig` sections. If you are using those fields for any of the new build-in integrations,
+you can replace your `extraConfig` to the new `integrations` feature.
```
BeverlyJaneJ, line 303: Do you mean built-in integrations?
Not sure what this means. Does it mean if you used extraConfig before, you should now replace it with integrations?

BeverlyJaneJ (approved):

BeverlyJaneJ (approved):

charts/k8s-monitoring/docs/Migration.md
petewall: In v2, we have included the configs for mysql, etcd, cert-manager. What I'm trying to say is if you have used `extraConfig` to add those metrics, you don't need to use extraConfig anymore, you can use the new `integrations` feature to pick up that data instead.

BeverlyJaneJ: Does this capture what you mean? I guess my question is if I've configured with extraConfig, do I have to change it? If so, we should probably say that.

Integrations are a new feature in v2.0 that allow you to enable and configure additional data sources. This also
includes the Alloy metrics that were previously part of `v1`.

Previously in v1, the configurations for MySQL, etcd, and cert-manager had to be defined in the `extraConfig` and `logs.extraConfig` sections. Now the configurations for MySQL, etcd, and cert-manager are built in. 
For v2.x.x, define the configuration for these services in the `integrations` feature.

petewall: that's a good call out: Just tell me if I have to do something

Let me rework a bit and update

BeverlyJaneJ (approved):
woot!

BeverlyJaneJ (approved):
woot!

PR #1175, opened 2025-01-28
set service_name label for podLogs

rlankfo:
This PR sets `service_name`, `service_namespace`, `deployment_environment`, and `deployment_environment_name` labels on Loki pod logs. This works when sent to loki destinations. I'll follow this PR up with updates for pod logs sent to OTLP destinations.

The labels are set according to the following rules:
- `service_name` label set by taking the first non-empty value from these Kubernetes metadata fields (in order): the pod annotation `resource.opentelemetry.io/service.name`, the pod label `app.kubernetes.io/name`, the pod’s name, or the container’s name.
- `service_namespace` set by `resource.opentelemetry.io/service.namespace` pod annotation.
- `deployment_environment` set by `resource.opentelemetry.io/deployment.environment` pod annotation.
- `deployment_environment_name` set by `resource.opentelemetry.io/deployment.environment.name` pod annotation.

If not set, Loki will attempt to set a default value according to these rules: https://grafana.com/docs/loki/latest/get-started/labels/#default-labels-for-all-users

### DIFF
--- a/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
+++ b/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
@@ -61,6 +61,28 @@ discovery.relabel "filtered_pods" {
     regex = "__meta_kubernetes_pod_annotation_(.+)"
   }
 
+  // explicitly set service_name. if not set, loki will automatically try to populate a default.
+  // see https://grafana.com/docs/loki/latest/get-started/labels/#default-labels-for-all-users
+  //
+  // choose the first value found from the following ordered list:
+  // - pod.annotation[resource.opentelemetry.io/service.name]
+  // - pod.label[app.kubernetes.io/name]
+  // - k8s.pod.name
+  // - k8s.container.name
+  rule {
+    action = "replace"
+    source_labels = [
+      "__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name",
+      "__meta_kubernetes_pod_label_app_kubernetes_io_name",
+      "__meta_kubernetes_pod_name",
+      "__meta_kubernetes_pod_container_name",
+    ]
+    separator = ";"
+    regex = "^(?:;*)?([^;]+).*$"
+    replacement = "$1"
+    target_label = "service_name"
+  }
+
 {{- if .Values.extraDiscoveryRules }}
 {{ .Values.extraDiscoveryRules | indent 2 }}
 {{- end }}
--- a/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
+++ b/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
@@ -83,6 +83,26 @@ discovery.relabel "filtered_pods" {
     target_label = "service_name"
   }
 
+  // set service_namespace
+  rule {
+    action = "replace"
+    source_labels = ["__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_namespace"]
+    target_label = "service_namespace"
+  }
+
+  // set deployment_environment and deployment_environment_name
+  rule {
+    action = "replace"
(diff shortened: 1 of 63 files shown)


petewall (approved):
LGTM. Run `make build` inside `charts/k8s-monitoring` to update the generated files.

petewall (approved):

charts/k8s-monitoring/tests/platform/otlp-gateway/values.yaml
```diff
@@ -25,5 +30,11 @@ integrations:
...
           app.kubernetes.io/name: ["alloy-metrics"]
 
+podLogs:
+  enabled: true
+
 alloy-metrics:
   enabled: true
+
+alloy-logs:
+  enabled: true
```
rlankfo, line 40:
```suggestion
  enabled: true
```

charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
```diff
@@ -61,6 +61,44 @@ discovery.relabel "filtered_pods" {
...
+  rule {
+    action = "replace"
+    source_labels = ["__meta_kubernetes_pod_label_app_kubernetes_io_name"]
+    target_label = "service_name"
+  }
+  rule {
+    action = "replace"
+    source_labels = ["__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name"]
+    target_label = "service_name"
+  }
```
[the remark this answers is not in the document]
rlankfo: Thanks @bentonam! I thought I could probably make something like this work and decided to come back to it. This is helpful.

Regarding the actual source labels, the intention is to follow suit with what the opentelemetry operator does in terms of calculating `service.name`: https://github.com/open-telemetry/opentelemetry-operator?tab=readme-ov-file#how-servicename-is-calculated.

Beyla and others have already gone this route so we are planning to do the same. From here, if `service_name` remains unset, Loki will attempt to create this label as described in the docs here: https://grafana.com/docs/loki/latest/get-started/labels/#default-labels-for-all-users

charts/k8s-monitoring-test/Chart.lock
```diff
@@ -1,3 +1,9 @@
-dependencies: []
-digest: sha256:643d5437104296e21d906ecb15b2c96ad278f20cfc4af53b12bb6069bd853726
-generated: "2024-10-10T14:33:15.167541-05:00"
+dependencies:
+- name: hello-kubernetes
+  repository: file://charts/hello-kubernetes
+  version: 1.0.0
+- name: hello-kubernetes
+  repository: file://charts/hello-kubernetes
+  version: 1.0.0
```
rlankfo, line 7: The intention of this was to have some test pods where we could add labels, etc. 

After some thought on this, it makes more sense to stand up a couple of `hello-kubernetes` pods in the deployment manifest of the specific tests that might need them. This way, these won't always get deployed with the test chart which should be a much cleaner approach.

charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
```diff
@@ -61,6 +61,28 @@ discovery.relabel "filtered_pods" {
...
+      "__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name",
+      "__meta_kubernetes_pod_label_app_kubernetes_io_name",
+      "__meta_kubernetes_pod_name",
+      "__meta_kubernetes_pod_container_name",
+    ]
+    separator = ";"
+    regex = "^(?:;*)?([^;]+).*$"
+    replacement = "$1"
+    target_label = "service_name"
+  }
```
rlankfo, line 84: @cyrille-leclerc do you mind taking a look at this?

I can't include other values used by the opentelemetry operator (statefulset, deployment, etc.) because they are not currently supported by prometheus kubernetes service discovery.

cyrille-leclerc: Looks good to me thanks! I did a quick test on https://regex101.com/ and it seems to work but I didn't test in real life.

petewall (approved):

PR #1190, opened 2025-01-31
fix dead docs links in prometheus operator chart

mbaykara:


petewall:
Ooh, thanks for catching this!
Can you make the changes in `charts/k8s-monitoring/charts/feature-prometheus-operator-objects/README.md.gotmpl`? That's the true source of the README.md file.

mbaykara:
@petewall Thank you for pointing that out! I’ve made the changes accordingly.

petewall (approved):

PR #1205, opened 2025-02-05
Rlankfo/batch size

rlankfo:
Batch size is the "number of spans, metric data points, or log records after which a batch will be sent regardless of the timeout". See https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md#batch-processor

This PR sets default batch size for feature-application-observability chart to 8192.


petewall (approved):

petewall:
Adding @bentonam , because I believe it was your PS configs that influenced the default batch size.

bentonam (approved):
LGTM

PR #1211, opened 2025-02-07
chore(templates-destinations): handle if destination protocol is not set to grpc or http

kundkingan:
We encountered an unexpected error due to mistakenly setting "otelhttp" as the protocol, which prevented the Alloy receiver from starting due to a configuration issue in config.alloy


petewall:
I think this is an excellent change.
What I'd like to do is set it in the values.schema.json for the chart.

kundkingan:
> I think this is an excellent change. What I'd like to do is set it in the values.schema.json for the chart.

Do you mean this part in the values.schema.json?
```
                "protocol": {
                    "type": "string",
                    "enum": [
                        "grpc",
                        "http"
                    ]
                },

```

PR #1222, opened 2025-02-10
Fix Pod log annotation and label assignment

petewall:
Before, it was using the implicit `(.*)` regex, which means it would unset labels if the annotation was not defined.
Because we have `k8s.grafana.com/logs.job` ==> `job`, it means we always removed the job label.

Added an integration test to validate the change (it reproduced the failure initially)


skl (approved):

stefanandres:
Thanks for the quick fix! 🙂

PR #1235, opened 2025-02-13
fix: receiver extraService selector

kinolaev:
For now `app.kubernetes.io/name` is set to `k8s-monitoring` instead of `alloy-receiver`, so no alloy-receiver's pods are matched by extraService's selector.


petewall (approved):
Amazing. Thank you for catching and fixing this!

PR #1237, opened 2025-02-13
Add migration doc link to v2 chart README

mrethers:


petewall (approved):

petewall:
Can you make the same change in `charts/k8s-monitoring/README.md.gotmpl`? That file is what generates the actual README.md file, so both should be updated at the same time.

petewall:
Perfect! Thanks!

PR #1238, opened 2025-02-13
Add Receivers block to App O11y in Values

burnettdev:
When leveraging the Application Observability functionality, you need to specify the receivers block so Alloy knows what is enabled (i.e. otlp, grpc etc).

The current Values file does not have receivers: {}, and you only find out it's needed via the Validation when running Helm.

This PR simply just adds it in with no config, but links back to the documentation. There's also a fix on the type of what the function is for (currently states pod logs, now updated to enabling the receiving of application observability).


petewall:
This is super intuitive and helpful. Thanks! One last thing is that the README.md file is generated from the values.yaml file.
Can you run `make README.md` from inside of `charts/k8s-monitoring`, which should build the doc with the addition?

Thanks!

petewall (approved):

PR #1239, opened 2025-02-13
feat: add pod_association sources

kinolaev:
Using `k8s.pod.ip` and `k8s.pod.uid` resource attributes might be a better defaults for `k8sattributes` processor. Especially because `internalTrafficPolicy` on alloy-receiver's service is set to `Cluster` by default, so there is a chance that alloy would see a node ip address instead of pod's one and wouldn't be able to add k8s attributes.

This sources are also added by default in [Opentelemetry Collector chart](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/opentelemetry-collector-0.116.0/charts/opentelemetry-collector/templates/_config.tpl#L215).


rlankfo (approved):
lgtm

PR #1245, opened 2025-02-14
Add the ability to set filters on otlp destinations

petewall:


charts/k8s-monitoring/destinations/otlp-values.yaml
```diff
@@ -229,6 +229,24 @@ processors:
...
+    enabled: false
+    # -- Metric filters
+    # @section -- Filter Processor
+    metrics:
+      metric: []
+      datapoint: []
+    # -- Log filters
+    # @section -- Filter Processor
+    logs:
+      logRecord: []
```
rlankfo, line 243:
```suggestion
      log_record: []
```

we should match what we have in `feature-application-observability/values.yaml`

charts/k8s-monitoring/destinations/otlp-values.yaml
```diff
@@ -229,6 +229,24 @@ processors:
       resource: []
       span: []
       spanevent: []
+
+  filter:
```
rlankfo, line 233:
```suggestion
  filters:
```
we should match what we have in feature-application-observability/values.yaml

rlankfo (approved):
lgtm. left a couple suggestions

charts/k8s-monitoring/destinations/otlp-values.yaml
petewall: Yeah... but I want to standardize somewhere and IMO we've got way more camel case than snake case

charts/k8s-monitoring/destinations/otlp-values.yaml
petewall: This one I agree with.

PR #1246, opened 2025-02-15
set resource attributes on otlp destination telemetry

rlankfo:
- k8s.cluster.name / cluster
- service.name / service.namespace
- deployment.environment / deployment.environment.name


petewall (approved):
🚀

PR #1248, opened 2025-02-18
Remove redundant cadvisor relabel rules in alloyModule

PaulFarver:
Thanks for maintaining this awesome chart! It has saved us a lot of trouble ❤️ 

We currently have trouble with our openshift clusters, where cadvisor does not populate the `image` label on `container_network_.*` metrics for some reason. We saw the same problem with v1 of the chart, but were able to work around it with `dropEmptyImageLabels: false` and an `extraMetricProcessingRules`

The options in the k8s-monitoring chart v2:

`cadvisor.metricsTuning.dropEmptyContainerLabels`
`cadvisor.metricsTuning.dropEmptyImageLabels`
`cadvisor.metricsTuning.normalizeUnnecessaryLabels`
`cadvisor.metricsTuning.keepPhysicalFilesystemDevices`
`cadvisor.metricsTuning.keepPhysicalNetworkDevices`

currently do not seem to have an impact on how alloy behaves in reality, because these rules are always included through the `-alloy-module-kubernetes` configmap. 

Looking at our output k8s-monitoring-alloy-metrics configmap, we have this:

```yaml
data:
  config.alloy: |-
    ...
    // Feature: Cluster Metrics
    declare "cluster_metrics" {
      argument "metrics_destinations" {
        comment = "Must be a list of metric destinations where collected metrics should be forwarded to"
      }

      remote.kubernetes.configmap "kubernetes" {
        name = "k8s-monitoring-alloy-module-kubernetes"
        namespace = "k8s-monitoring"
      }

      import.string "kubernetes" {
        content = remote.kubernetes.configmap.kubernetes.data["core_metrics.alloy"]
      }
    ...
```

The k8s-monitoring-alloy-module-kubernetes includes the drop rules, but we cannot disable these

`charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_cadvisor.alloy.tpl` and `charts/k8s-monitoring/alloyModules/modules/kubernetes/core/metrics.alloy` seem to have the same set of rules, but `charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_cadvisor.alloy.tpl` is configurable


petewall:
Thanks, but in #1257, I'm actually breaking the use of `alloyModules` for cadvisor. That PR will have a more permanent and cleaner solution.

PaulFarver:
@petewall Cool. Looking forward to it 👍 
I'll close this PR, then

petewall:
This should be doable now with the 2.0.12 release!

PR #1254, opened 2025-02-19
fix: pod_association

kinolaev:
I'm sorry, I made a mistake in https://github.com/grafana/k8s-monitoring-helm/pull/1239, there are must be three `pod_association` instead of three `source` inside one `pod_association`. All `source` blocks has to match for the pod to be associated with a signal but we want it to be associated when one of the conditions is met (as in [Opentelemetry Collector chart](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/opentelemetry-collector-0.116.0/charts/opentelemetry-collector/templates/_config.tpl#L215)).


kinolaev:
Hello @petewall and @rlankfo, sorry for pinging you but I believe it is very important fix, because my previous PR (part of release 2.0.11) could break k8sattributes processor for many users

rlankfo (approved):
thanks for catching this!

PR #1257, opened 2025-02-19
Make cadvisor, kubelet, and kubelet resource use direct node access

petewall:
Also changes cluster metrics unit tests to use snapshots, which will be easier to compare and update.


charts/k8s-monitoring/charts/feature-cluster-metrics/tests/custom_rules_test.yaml
```diff
@@ -27,247 +27,6 @@ tests:
     asserts:
       - isKind:
           of: ConfigMap
-      - equal:
+      - matchSnapshot:
```
rlankfo, line 30: 😍

charts/k8s-monitoring/charts/feature-cluster-metrics/values.yaml
```diff
@@ -88,10 +99,21 @@ kubeletResource:
   # @section -- Kubelet Resources
   enabled: true
 
+  # -- How to access the node services, either direct (use node IP, requires nodes/metrics) or via proxy (requires nodes/proxy)
+  # @section -- Kubelet Resources
+  nodeAddressFormat: direct
```
rlankfo, line 104: just looking at this, it's not entirely clear to me. As a user, can I set this to "direct" or "proxy" ?

rlankfo (approved):
LGTM

charts/k8s-monitoring/charts/feature-cluster-metrics/values.yaml
petewall: Old:
```
# -- How to access the node services, either direct (use node IP, requires nodes/metrics) or via proxy (requires nodes/proxy)
```
New:
```
  # -- How to access the Kubelet, either "direct" (use node IP) or "proxy" (uses API Server)
```

PR #1284, opened 2025-02-28
Fix the ability to set scrape interval with an annotation.

petewall:
The trick is to use the __scrape_interval__ meta label, and check if the label was set by the annotation, then apply the defaults. It cannot be done in the prometheus.scrape component, because that would always override the annotation-set value.

This also adds the ability to set the scrape timeout by an annotation.


rlankfo (approved):
lgtm 🏄‍♂️

PR #1304, opened 2025-03-07
Add eks addon tests

petewall:
Signed-off-by: Pete Wall <pete.wall@grafana.com>

### DIFF
--- a/charts/k8s-monitoring/tests/platform/eks-addon-with-terraform/.gitignore
+++ b/charts/k8s-monitoring/tests/platform/eks-addon-with-terraform/.gitignore
@@ -1,0 +1,10 @@
+.norender
+.terraform
+.terraform.lock.hcl
+kubeconfig.yaml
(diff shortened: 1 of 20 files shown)


charts/k8s-monitoring/tests/platform/eks-addon-with-terraform/.gitignore
```diff
@@ -0,0 +1,10 @@
+.norender
```
skl, line 1: did you mean to include `.norender` in the gitignore? This is a `.no-render` file but I think you want to keep that, iiuc

skl (approved):

charts/k8s-monitoring/tests/platform/eks-addon-with-terraform/.gitignore
petewall: That can be removed. I'll do that.

PR #1315, opened 2025-03-12
Minor fixes:

petewall:
* Fix copy-paste error that put "cadvisor" in nearly all cluster metric sources
* Fix missing registry overrides for configreloader example
* Fix typo in private registry example check and make sure both examples are checked


charts/k8s-monitoring/charts/feature-cluster-metrics/values.yaml
```diff
@@ -206,7 +206,7 @@ cadvisor:
   # @section -- cadvisor
   scrapeInterval: ""
 
-  # -- Sets the max_cache_size for cadvisor prometheus.relabel component.
+  # -- Sets the max_cache_size for tbe cAdvisor prometheus.relabel component.
```
rlankfo, line 209:
```suggestion
  # -- Sets the max_cache_size for the cAdvisor prometheus.relabel component.
```

rlankfo (approved):

PR #1325, opened 2025-03-14
Update the fleet management example to use the password in the secret

petewall:
Don't define the secret in the env var


rlankfo (approved):
🚀

PR #1335, opened 2025-03-17
Properly build the list of targets when pods or services are disabled.

petewall:


charts/k8s-monitoring/charts/feature-annotation-autodiscovery/templates/_validation.tpl
```diff
@@ -0,0 +1,14 @@
+{{- define "feature.annotationAutodiscovery.validate" }}
+{{- if and (not .Values.pods.enabled) (not .Values.services.enabled) }}
```
skl, line 2: This was my first question, what if both are selected... so 💯 !

skl (approved):

PR #1337, opened 2025-03-18
Major action and dependabot updates

petewall:
1. Replace ReviewDog with a more basic Lint workflow that relies on the same Makefile targets we use on a workstation
2. Remove auto-approve workflows that weren't working
3. Remove updatecli and switch to beta helm dependabot


.github/workflows/auto-approve-dependabot.yml
skl, line 1: this one does actually work ([example](https://github.com/grafana/k8s-monitoring-helm/pull/1330)), the chart deps one does not. up to you if you keep this one or not (it doesn't auto merge, only auto approve)

skl (approved):
💯

PR #1362, opened 2025-03-24
Updating extra-configuration example to better explain naming syntax

electrominium:
The destination name and alloy prometheus remote write receiver names have different naming conventions. Updating the example to reflect that.


petewall:
I just updated the same example for another reason. You may need to rebase and build again. Also, the CLA is required, too!
Thanks!

electrominium:
@petewall I have already signed the CLA but not sure why it does not show up.

petewall:
Looks like one last generated file to update. If you cd into charts/k8s-monitoring and run make examples, it should rebuild this.

electrominium:
Done

PR #1366, opened 2025-03-24
Don't validate credentials configured as external

### LINKED ISSUE #1364: Support Referencing External Secret for Credentials
To make deployment easier when using GitOps tools, it's often helpful to define credentials through something external then reference them within the Helm values. This doesn't seem to currently be possible for the destination or remote management credentials within this Chart.

I'd expect this to look something like the below.

```yaml
destinations:
  - name: grafana-cloud-metrics
    type: prometheus
    url: https://prometheus-prod-05-gb-south-0.grafana.net/api/prom/push
    auth:
      type: basic
      existingSecret: <name>
```

This allows for keeping the Alloy deployment within GitOps but removes the need to commit plaintext credentials.

Apologies if this is already possible, but I couldn't see a way to do it when tracing through the Helm Chart. Several of my attempts at bodging it ran afoul of #1005.

thebeanogamer:
It's currently not possible to reference external secrets for credentials as the linter is trying to validate parameters that you've deliberately not included.

This change keeps the existing validation around keys for secrets managed by this chart but disables it when referencing an external secret.

Resolves #1364

### DIFF
--- a/charts/k8s-monitoring/templates/destinations/_destination_validations.tpl
+++ b/charts/k8s-monitoring/templates/destinations/_destination_validations.tpl
@@ -64,11 +64,13 @@
     {{- end }}
 
     {{- if eq (include "secrets.authType" $destination) "basic" }}
-      {{- if eq (include "secrets.usesSecret" (dict "object" $destination "key" "auth.username")) "false" }}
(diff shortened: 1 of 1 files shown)


thebeanogamer:
:facepalm: this is already handled, an unrelated issue with my test environment made me miss that

PR #1367, opened 2025-03-25
Add a test and example of using stage.metrics.

petewall:
Add a test and example of using stage.metrics.
Also add an Alloy integration test.


rlankfo (approved):

charts/k8s-monitoring/docs/examples/log-metrics/README.md
```diff
@@ -0,0 +1,59 @@
...
+    url: http://prometheus.prometheus.svc:9090/api/v1/write
+
+  - name: loki
+    type: loki
+    url: http://loki.loki.svc:3100/loki/api/v1/push
+
+podLogs:
+  enabled: true
+
+  extraLogProcessingStages: |-
```
mvadu, line 31: In https://github.com/grafana/k8s-monitoring-helm/blob/25c2df7b95c35f902242b2e9314856b99420b99a/charts/k8s-monitoring/charts/feature-node-logs/templates/_module.alloy.tpl#L150 there is no pod filter. Does that mean if we have multiple stages, each stage gets applied to every single pod log being scraped? Isn't it ineffective?

PR #1368, opened 2025-03-25
Fix label of prometheus.operator.probes component

ndrpnt:
Small typo fix. Should I update the corresponding test snapshots ?


petewall:
Thank you!
The GH Action tests will tell us if we need to update the unit tests.

petewall:
Yeah, please update the test snapshots.

petewall (approved):

PR #1381, opened 2025-03-31
Make the cluster labels configurable per destination

petewall:


rlankfo (approved):

alextricity25:
I think this broke the cluster name labels for loki. I have this setting in my values.yaml:
```
    'cluster': {
      name: 'my-cluster-label',
    },
```

And after upgrading to version 2.0.23 this morning my loki logs no longer have this label. They now have the label `cluster: loki`, which I presume is the default for this destination.

Even after adding the `clusterLabels` to my `loki` destination, the cluster label still appears as `loki`, rather than `my-cluster-label`. 

```
      {
        name: 'localLoki',
        type: 'loki',
        clusterLabels: ['cluster'],
        url: <my-url>
        logs: {
          enabled: true,
        },
      },
```

petewall:
I'm guessing what happened is that this fixed an issue where the cluster label was being overwritten. Loki is setting its own `cluster` label and the chart was mistakenly overwriting it.
I'd suggest utilizing the `k8s.cluster.name` label, or we could introduce a relabeling rule to drop Loki's cluster label

PR #1389, opened 2025-03-31
Add jobLabel to annotation autodiscovery

rubensf:
It was missing.

### DIFF
--- a/charts/k8s-monitoring/charts/feature-annotation-autodiscovery/README.md
+++ b/charts/k8s-monitoring/charts/feature-annotation-autodiscovery/README.md
@@ -72,6 +72,7 @@ Be sure perform actual integration testing in a live environment in the main [k8
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | bearerToken | object | `{"enabled":true,"token":"/var/run/secrets/kubernetes.io/serviceaccount/token"}` | Sets bearer_token_file line in the prometheus.scrape annotation_autodiscovery. |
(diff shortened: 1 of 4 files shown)


rubensf:
Sorry I missed your notification here - we do have that `jobLabel` as an option across a bunch of other spots, such as everything under `clusterMetrics`. Why not add that as an option here?

Strictly, we already have a label that defaults to `k8s.grafana.com/job` under `annotations.job` value: https://github.com/grafana/k8s-monitoring-helm/blob/9536973ed2454f5454e75e047f8e96fdc5cfeb2f/charts/k8s-monitoring/charts/feature-annotation-autodiscovery/values.yaml#L32

rubensf:
Just updated the PR to make it optional :)

rubensf:
ping @petewall

rubensf:
ping @petewall @nevermind89x @skl . Sorry for the spam!

petewall:
So, after all that, what's the course of action? If you want to take this on, find the Alloy rule where we set `job` from the `k8s.grafana.com/job` annotation. Then make copies of that rule for the other options, but checking if job was not yet set. e.g.:

```
rule {
  source_labels = ["job", "__meta_kubernetes_pod_label_app_kubernetes_io_name"]
  regex = ";(.+)"
  target_label = "job"
}
```

This will only set `job` to `app.kubernetes.io/name` if the job label is not currently set (because of nothing before the default separator `;`).

petewall:
closing in favor of #1717

PR #1392, opened 2025-04-01
The proxyURL is not honored by the k8s-monitoring helm template.

darlingtld:
`proxyURL` is specified in the [schema](https://github.com/grafana/k8s-monitoring-helm/blob/8075584af7292ec5248877aede16bf28edef9044/charts/k8s-monitoring/values.schema.json), but the field is not honored when specified. Hence fixed the bug.

### DIFF
--- a/charts/k8s-monitoring/values.yaml
+++ b/charts/k8s-monitoring/values.yaml
@@ -431,7 +431,7 @@ alloy-singleton:
     url: ""
 
     # -- The proxy URL to use of the remote config server.
-    # @section -- Collectors - Alloy Metrics
+    # @section -- Collectors - Alloy Singleton
     proxyURL: ""
 
     auth:
@@ -555,7 +555,7 @@ alloy-logs:
     url: ""
 
     # -- The proxy URL to use of the remote config server.
-    # @section -- Collectors - Alloy Metrics
+    # @section -- Collectors - Alloy Logs
     proxyURL: ""
 
     auth:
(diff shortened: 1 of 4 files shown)


petewall:
Currently, we don't have `proxyURL` set in the values.yaml for remoteConfig. All the existing entries in the schema are for the destinations.  We can/should add this to the values file for `alloy-___.remoteConfig`.

darlingtld:
> Currently, we don't have `proxyURL` set in the values.yaml for remoteConfig. All the existing entries in the schema are for the destinations. We can/should add this to the values file for `alloy-___.remoteConfig`.

Hi @petewall Thanks for the comments. I added the `proxyURL` in the schema file and `alloy-*.remoteConfig` as well. Please review and suggest.

charts/k8s-monitoring/values.yaml
```diff
@@ -813,6 +829,10 @@ alloy-profiles:
     # @section -- Collectors - Alloy Profiles
     url: ""
 
+    # -- The proxy URL to use of the remote config server.
+    # @section -- Collectors - Alloy Metrics
```
petewall, line 833: This should be Alloy Profiles.
I think the same applies to other sections, too.

Then, rebuild README.md.

petewall (changes requested):

charts/k8s-monitoring/values.yaml
darlingtld: Thanks. The fix has been pushed.

petewall:
Great! Thanks for the changes!

PR #1428, opened 2025-04-14
Drop deployemnt label for clusterMetrics when no component is managed by helmchart

transacid:
when all clusterMetrics components are managed outside of this helm chart (`deploy: false`), the `self-reporting-metric.prom` snippet includes a label value `deployments="%!s(<nil>)"`.\
This PR checks whether there are deployments managed by this chart and only then creates the label.

### DIFF
--- a/charts/k8s-monitoring/tests/integration/auth/.rendered/output.yaml
+++ b/charts/k8s-monitoring/tests/integration/auth/.rendered/output.yaml
@@ -788,7 +788,7 @@ data:
     grafana_kubernetes_monitoring_build_info{version="2.0.23", namespace="default"} 1
     # HELP grafana_kubernetes_monitoring_feature_info A metric to report the enabled features of the Kubernetes Monitoring Helm chart
     # TYPE grafana_kubernetes_monitoring_feature_info gauge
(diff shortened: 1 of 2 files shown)


petewall:
This is great. Can you update the generated files by running "make clean build test" at the top level?

charts/k8s-monitoring/tests/integration/auth/.rendered/output.yaml
```diff
@@ -788,7 +788,7 @@ data:
     grafana_kubernetes_monitoring_build_info{version="2.0.23", namespace="default"} 1
     # HELP grafana_kubernetes_monitoring_feature_info A metric to report the enabled features of the Kubernetes Monitoring Helm chart
     # TYPE grafana_kubernetes_monitoring_feature_info gauge
-    grafana_kubernetes_monitoring_feature_info{deployments="%!s(<nil>)", feature="clusterMetrics", sources="kubelet", version="1.0.0"} 1
```
petewall, line 791: Love that the fix actually does something in our existing examples!

petewall (approved):
This is great! Thank you!

PR #1431, opened 2025-04-16
Add a test to check upgrades at the patch level

petewall:
* Install first at x.y.(z-1)
* Test for data (which blocks until data is available)
* Upgrade to x.y.z
* Test again

The chart sets a `chart_version` label to all metrics and logs with the current Helm chart version. This is how we ensure the upgrade was successful.

Successful run: https://github.com/grafana/k8s-monitoring-helm/actions/runs/14484180399/job/40626674396


rlankfo (approved):
LGTM although it'd be great to also test an OTLP destination at some point here.

PR #1434, opened 2025-04-16
Fix infinite diff in openshift alloy logs scc

PaulFarver:
The current scc for alloy logs in openshift renders the scc with hostpath first in the array. But it seems openshift reorders this upon apply. But this causes gitops systems to detect a diff and causes a lot of unecessary syncs.

This is similar to how the scc looks for `charts/k8s-monitoring-v1/templates/platform_specific/openshift/alloy-scc.yaml`, so I assume this was the case there as well.

image

### DIFF
--- a/charts/k8s-monitoring/docs/examples/platforms/openshift/output.yaml
+++ b/charts/k8s-monitoring/docs/examples/platforms/openshift/output.yaml
@@ -2776,9 +2776,9 @@ supplementalGroups:
   type: RunAsAny
 users: []
 volumes:
-- hostPath
 - configMap
 - emptyDir
+- hostPath
 - projected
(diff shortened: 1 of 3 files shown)


petewall (changes requested):
Makes sense. Can you make the same change in these generated files:

* `charts/k8s-monitoring/docs/examples/platforms/openshift/output.yaml`
* `charts/k8s-monitoring/tests/platform/openshift/.rendered/output.yaml`

PaulFarver:
@petewall thanks for the quick reaction :)
I couldn't figure out how to generate the files, so I updated them manually. Hope that is ok.

petewall (approved):

PR #1450, opened 2025-04-23
Fix the doc for batch processor size

gouthamve:
The current docs make it seem like the limit is in bytes. This is misleading. I copied over the original text from here: https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md


petewall (changes requested):
This is very helpful! Can you make the same change inside of `charts/k8s-monitoring/destinations/otlp-values.yaml` as well? This README.md is generated from that file.

You can also modify `charts/k8s-monitoring/destinations/otlp-values.yaml`, then run `make README.md` from inside the `charts/k8s-monitoring` directory.

petewall:
Fixing in #1499

PR #1466, opened 2025-04-28
feat: look at otel annotations to rectify resource attributes

cedricziel:


ChristianCiach:
This should already work due to the configuration of the k8sattributes processor in the same file:

```yaml
    annotation {
      key_regex = "resource.opentelemetry.io/(.*)"
      tag_name  = "$1"
      from      = "pod"
    }
```

PR #1500, opened 2025-05-03
Go back to including only destinations that are used

petewall:
But also allow for specify destinations that should be included anyway


rlankfo (approved):
🚀

PR #1510, opened 2025-05-06
feat: Adding in Service Graph Metrics to App O11y Feature

Imshelledin21:
Adds in the ability to generate Service Graph Metrics in the Application Observability Feature utilizing the [otelcol.connector.servicegraph](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.connector.servicegraph) component.


petewall:
Service Graph Metrics generation was added in 3.1

PR #1519, opened 2025-05-08
Cursor rules file for k8s-monitoring v2

daveram:
Cursor rules file for use with cursor.ai to provide context and guidance to AI models to relevant docs/files when being asked questions about using the k8s-monitoring v2 helm chart


petewall (commented):
This is awesome. Do we want to also include the values.yaml files for:
* feature subcharts (e.g. `charts/k8s-monitoring/feature-cluster-metrics/values.yaml`)
* destination values (e.g. `charts/k8s-monitoring/destinations/prometheus-values.yaml`)
* integrations (e.g. `charts/k8s-monitoring/feature-integrations/integrations/loki-values.yaml`)?

I don't know if it's smart enough to know that otlp destinations go inside the `destinations` section of the main values file, but that could be very helpful.

petewall (approved):

daveram:
Unfortunately no good way to glob the values and readme files, so we have to list them out

petewall (approved):

PR #1524, opened 2025-05-08
Add step in the release pipeline to package the helm chart as an OCI …

petewall:
…image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,3 +129,13 @@ jobs:
           CR_TOKEN: ${{ secrets.GH_BOT_ACCESS_TOKEN }}
         run: |
           "${BIN_PATH}/cr" index --index-path "${INDEX_PATH}" --push
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772  # v3.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push chart as OCI image to GitHub Container Registry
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -129,3 +129,13 @@ jobs:
           CR_TOKEN: ${{ secrets.GH_BOT_ACCESS_TOKEN }}
         run: |
           "${BIN_PATH}/cr" index --index-path "${INDEX_PATH}" --push
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772  # v3.4.0
```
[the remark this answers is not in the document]
petewall: This is a bad idea, since tags can be hijacked, but commit sha's cannot.

.github/workflows/release.yml
```diff
@@ -129,3 +129,13 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
dat-intellis, line 138: @petewall > issue: If I remember correctly you need to have 'permissions: write:packages' for this token to be populated with the appropriate access rights. 

This is probably also the reason why the release pipelines have been failing since this was introduced: https://github.com/grafana/k8s-monitoring-helm/actions/workflows/release.yml

I'm a bit fuzzy on this, but you _might_ have to connect the package post-publishing as a confirm-once action.

petewall: Thanks for this. Yeah, I've been slow to try and fix it because the rest of the release process has been working as designed.

I'll try and address this during this change: https://github.com/grafana/k8s-monitoring-helm/issues/1574

PR #1541, opened 2025-05-13
fix: indentation on autoinstrumentaiton example

ampc:
Indentation issue on routes section


petewall:
This will rebuild some generated files. Can you go into `charts/k8s-monitoring` and run `make examples`?

PR #1542, opened 2025-05-14
Move `pod` and `k8s.pod.name` to structured metadata

petewall:


rlankfo (approved):
LGTM 🏄

charts/k8s-monitoring/charts/feature-pod-logs/values.yaml
```diff
@@ -127,15 +125,16 @@ labelsToKeep:
...
 # Example:
 # structuredMetadata:
 #   component: component
 #   kind: kind
 #   name: name
 # @section -- Processing settings
-structuredMetadata: {}
+structuredMetadata:
+  k8s.pod.name: k8s.pod.name
+  pod: pod
```
ChristianCiach, line 137: Is there a reason for using dots instead of underscores in the keys?

It's confusing that I have to override the default `labels` using underscores but I have to use dots for overriding the default `structuredMetadata`:

```yaml
podLogs:
  labels:
    app_kubernetes_io_name: ~
  structuredMetadata:
    k8s.pod.name: ~
```

When I change `k8s.pod.name:` to `k8s_pod_name:`, it won't override the default and the same mapping will be rendered twice in `stage.structured_metadata {...}`. Similarly, I cannot use dots in the `labels` configuration, because the helm chart doesn't escape the keys in `labels:`, but it will escape the keys in `structuredMetadata:`.

pschiffe:
For anyone who's wondering why you cannot use pod name in dashboard variable in Grafana anymore - structured metadata is not supported: https://github.com/grafana/grafana/issues/84909

PR #1547, opened 2025-05-14
Remove validation for clustering with volumes.

petewall:
It's not actually a functional check and has time when it's important


patrickeasters:
Thanks @petewall! Can confirm this works and passes validations in my test cluster with these values:
```yaml
nodeLogs:
  enabled: true
podLogs:
  enabled: true
alloy-logs:
  enabled: true
  alloy:
    clustering:
      enabled: true
    mounts:
      varlog: true
      dockercontainers: false
```

PR #1552, opened 2025-05-15
Update otlp.md

Utwo:
Fix typo in destinations docs

### DIFF
--- a/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
+++ b/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
@@ -276,7 +276,7 @@ otelcol.processor.filter {{ include "helper.alloy_name" .name | quote }} {
 {{- end }}
 {{- if and .logs.enabled .processors.filters.logs.logRecord }}
   logs {
-    logRecord = [
+    log_record = [
 {{- range $filter := .processors.filters.logs.logRecord }}
 {{ $filter | quote | indent 6 }},
 {{- end }}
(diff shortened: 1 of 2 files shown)


Utwo:
That was extremely fast! In the .tpl I can see it like this. https://github.com/grafana/k8s-monitoring-helm/blob/af8f0c0a762157044003b1ff01a1478f29838e23/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl#L277

Also, I could only make it work with `log_record` on my side.

petewall:
Ah! Then it's a bug in the code. Could you instead fix the tpl file? I'd rather have consistency in our naming (as much as possible). Both on that line and the one a few lines later?

charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -274,10 +274,10 @@ otelcol.processor.filter {{ include "helper.alloy_name" .name | quote }} {
 {{- end }}
   }
 {{- end }}
-{{- if and .logs.enabled .processors.filters.logs.log_record }}
+{{- if and .logs.enabled .processors.filters.logs.logRecord }}
   logs {
-    log_record = [
-{{- range $filter := .processors.filters.logs.log_record }}
+    logRecord = [
```
petewall, line 279:
```suggestion
    log_record = [
```
This one has to stay, since it corresponds to this Alloy config:
https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.filter/#logs-block

Utwo:
Thank you!

petewall (approved):

Utwo:
I think you will need to merge it. I don't have permissions.

PR #1586, opened 2025-05-21
Update istio examples to show how to get alloy-receiver working

petewall:
also add the includeMetdata option for otlp receivers, even though it's not required for this


rlankfo (approved):
LGTM

PR #1604, opened 2025-05-24
load balanced service graph metrics

rlankfo:
This PR adds a new dynamic alloy statefulset to generate service graphs if configured for a traces destination. 

When scaling service graphs, it's important to distinguish which alloy instance generated the metrics, otherwise it's possible for collisions that will result in duplicate samples, out of order sample errors, etc. This also creates irregularities with service graph counters such as an unexpected sawtooth pattern. 

NOTE: This is remedied in this PR for ***prometheus destinations only!*** by adding a `collector_id` label at remote write time. This is based on the recommendations from [our documentation](https://grafana.com/docs/grafana-cloud/monitor-applications/application-observability/collector/sampling/scale/#scale-alloy-tail-sampling): 
In order to differentiate series generated by different collectors on the second layer we recommend adding additional label “collector_id”. Cardinality issues due to “collector_id” labels can be solved using [Adaptive Metrics](https://grafana.com/docs/grafana-cloud/monitor-applications/application-observability/setup/scaling/adaptive-metrics/).

OTLP destinations are still supported at this time but the `collector_id` label will not be added. Currently, even if we added this as a resource attribute to the metrics, the OTLP endpoint in Grafana Cloud would not promote this from an attribute to a label.

### DIFF
--- a/charts/k8s-monitoring/tests/integration/service-graph-metrics/deployments/query-test.yaml
+++ b/charts/k8s-monitoring/tests/integration/service-graph-metrics/deployments/query-test.yaml
@@ -41,7 +41,7 @@ spec:
             expect:
               value: 3
               operator: ==
-          - query: 'rate(traces_service_graph_request_total{cluster="service-graph-metrics-test", collector_id="k8smon-localtempo-servicegraph-alloy-0"}[5m]) > 0'
+          - query: 'rate(traces_service_graph_request_total{cluster="service-graph-metrics-test", collector_id="k8smon-localtempo-servicegraph-0"}[5m]) > 0'
             type: promql
-          - query: 'rate(traces_service_graph_request_total{cluster="service-graph-metrics-test", collector_id="k8smon-localtempo-servicegraph-alloy-1"}[5m]) > 0'
+          - query: 'rate(traces_service_graph_request_total{cluster="service-graph-metrics-test", collector_id="k8smon-localtempo-servicegraph-1"}[5m]) > 0'
             type: promql


charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -293,6 +305,25 @@ otelcol.exporter.loadbalancing {{ include "helper.alloy_name" .name | quote }} {
         }
       }
     }
+{{- end }}
+
+{{- if .processors.serviceGraphMetrics.enabled }}
+otelcol.exporter.loadbalancing {{ printf "%s_servicegraph" (include "helper.alloy_name" .name) | quote }} {
+  resolver {
+    kubernetes {
+      service = "{{ include "helper.k8s_name" (printf "%s-%s-servicegraph-alloy" $.Release.Name .name) }}"
```
petewall, line 314: Should this match the full business that the tailSampler's `kubernetes { ... }` section gets?
That's essentially for handling:
* Names that are too long
* `nameOverride` and `fullnameOverride` settings

charts/k8s-monitoring/tests/integration/service-graph-metrics/.rendered/output.yaml
```diff
@@ -0,0 +1,1940 @@
...
+      namespace = "default"
+    }
+    
+    // Destination: localTempo (otlp)
+    
+    otelcol.processor.attributes "localtempo" {
+      output {
+        traces = [otelcol.processor.transform.localtempo.input]
+      }
+    }
```
petewall, line 915: We should probably skip components if they no-op like this... not for now, just calling it out for a future improvement.

charts/k8s-monitoring/tests/integration/service-graph-metrics/.rendered/output.yaml
```diff
@@ -0,0 +1,1940 @@
...
+# Source: k8s-monitoring/templates/alloy-servicegraph.yaml
+apiVersion: collectors.grafana.com/v1alpha1
+kind: Alloy
+metadata:
+  name: k8smon-localtempo-servicegraph
+  namespace: default
+spec:
+  alloy:
+    configMap:
+      content: |2-
```
petewall, line 1240: I think we can fix this with the right `trim` command... I'll put another comment in the code that generates the config...

charts/k8s-monitoring/templates/alloy-servicegraph.yaml
```diff
@@ -0,0 +1,64 @@
...
+        {{- $destinationTargets = append $destinationTargets (printf "%s" $destinationTarget) }}
+      {{- end }}
+    {{- end }}
+
+    {{- $serviceGraphConfig := merge (dict "traces" $destination "metrics" $destinationTargets "name" "service_graph_metrics") $destination.processors.serviceGraphMetrics }}
+    {{- $serviceGraphTarget := include "servicegraph.connector.serviceGraphMetrics.alloy.target" $serviceGraphConfig }}
+    {{- $receiverConfig := merge (dict "traces" $serviceGraphTarget "name" "service_graph_receiver") $destination.processors.serviceGraphMetrics }}
+
+    {{- $alloyConfig := "" }}
+    {{- $alloyConfig = cat $alloyConfig ((include "receiver.otlp.alloy" $receiverConfig) | trim | nindent 0) }}
```
petewall, line 29: This might remove the `config: |2` thing we see when it's set.
```suggestion
    {{- $alloyConfig = cat $alloyConfig ((include "receiver.otlp.alloy" $receiverConfig) | trim | indent 0) }}
```

PR #1620, opened 2025-05-31
add optional tail sampling args

mbaykara:


petewall:
@rlankfo can you look at this?

charts/k8s-monitoring/destinations/otlp-values.yaml
```diff
@@ -238,6 +238,20 @@ processors:
     # @section -- Tail Sampling
     decisionWait: 15s
 
+    # -- The decision cache for the tail sampling.
+    # @section -- Tail Sampling
+    decisionCache: {}
+    #  sampledCacheSize: 60000
+    #  nonSampledCacheSize: 60000
```
rlankfo, line 245: Could we uncomment these and add a doc snippet so these show up in the table?

sampled_cache_size (default = 0): Configures amount of trace IDs to be kept in an LRU cache, persisting the "keep" decisions for traces that may have already been released from memory. By default, the size is 0 and the cache is inactive.
non_sampled_cache_size (default = 0) Configures amount of trace IDs to be kept in an LRU cache, persisting the "drop" decisions for traces that may have already been released from memory. By default, the size is 0 and the cache is inactive.

rlankfo (approved):
LGTM with one suggestion

charts/k8s-monitoring/destinations/otlp-values.yaml
mbaykara: I set the value 60k here due to default value of [num_traces](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.tail_sampling/#arguments) which is 50k.

PR #1622, opened 2025-05-31
feat: support raw string statements section in otlp destination

doubles-ss:
Due to the statement interpreter not supporting nested function calling, we can not provide value from environment variables to statements, but if we use `string.format` we can provide value from the environment into a valid statement, and this requires a raw string to do it.

https://github.com/grafana/alloy/issues/2054#issuecomment-2527963303


charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -65,16 +65,28 @@ otelcol.processor.transform {{ include "helper.alloy_name" .name | quote }} {
 {{- range $transform := .processors.transform.metrics.resource }}
 {{ $transform | quote | indent 6 }},
 {{- end }}
+{{- end }}
+{{- if .processors.transform.metrics.resourceFrom }}
+{{- range $transform := .processors.transform.metrics.resourceFrom }}
+{{ $transform | indent 6 }},
+{{- end }}
 {{- end }}
```
petewall, line 73: The range operator will no-op if the list is empty, so this can be simplified a bunch:
```suggestion
{{- range $transform := .processors.transform.metrics.resourceFrom }}
{{ $transform | indent 6 }},
{{- end }}
```

This applies to all of the other instances, too.

petewall (commented):
Thank you so much for your contribution. Just a small adjustment to the code.

Would you also be willing to find an update a example in the docs or an integration test to demonstrate how this works?

charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
doubles-ss: @petewall Should I remove `if statement` from the old processor (no "From" suffix) or just remove only the new one?

PR #1646, opened 2025-06-06
Update the release workflows to use the vault-based credentials

petewall:
Also delete lots of scripts and config files that are not used anymore.


rlankfo (approved):
🚀

PR #1688, opened 2025-06-21
fix: can't disable retry on failure otlp destination

doubles-ss:
Due to the default value in the template being set to true, when we input the attribute with false (bool) in `values.yaml` it will drop that attribute and cause it to use default values from the template (true).


petewall:
I think rather than changing the defaults, we should fix the reason that you can't change the value.

Let me check how this works currently and maybe I can recommend how to change things to get it working.

doubles-ss:
> 

I didn't change default values for deployment. I change logic to determine value from values.yaml file.
For a workaround, I can change false (bool) to "false" (string), but it doesn't make sense that we input a string for a boolean value. When we set a false (bool) value in values.yaml, it will drop that attribute; that's why we can't use the value from that attribute, and it always uses the default value in templates.

petewall:
Take a look at this PR and see if it meets your needs: https://github.com/grafana/k8s-monitoring-helm/pull/1700
I used unit tests to check that setting the field to false ensures that it works.

PR #1696, opened 2025-06-24
Properly drop log.file.path from the log context

petewall:
The `log.file.path` appears to be a log attribute, not a resource attribute, so we need to drop it in that context.

Also adds unit tests.


rlankfo (approved):
🚀

ChristianCiach:
Maybe you should make use of context inference: https://opentelemetry.io/blog/2025/ottl-contexts-just-got-easier/

Or maybe this is not yet available in Alloy?

PR #1727, opened 2025-07-03
Helm docs adjust after move

BeverlyJaneJ:
- Deleted content moved into main docs
- Updated Structure.md to provide guidance to users on structure of chart and where to look for values and docs
- In auto instrumentation, changed auto-instrument to zero-code

@petewall Need your help in finishing instructions for how to add a readme.md.


petewall:
This is really great! Let me get back to this after the holiday weekend.

BeverlyJaneJ:
Hi, @petewall , added links to docs

petewall:
The one I'm concerned about is `Images.md`, this tends to change each release.
If we can find a way to update grafana.com docs with a PR for that, that would be amazing...

PR #1738, opened 2025-07-07
Set `service.namespace` and `service.instance.id` in pod logs feature

petewall:


rlankfo (approved):

charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
```diff
@@ -75,6 +75,23 @@ discovery.relabel "filtered_pods" {
     target_label = "service_name"
   }
 
+  // explicitly set service_namespace.
+  //
+  // choose the first value found from the following ordered list:
+  // - pod.annotation[resource.opentelemetry.io/service.namespace]
+  // - pod.namespace
```
cyrille-leclerc, line 82: 👍 this is compliant with [Semantic conventions 1.35.0 / Non-normative / K8s attributes
Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/).

cyrille-leclerc:
My understanding is that this PR sets the pod log label:
```
service_name=firstNonNull(k8s.pod.annotation["resource.opentelemetry.io/service.namespace"], k8s.namespace.name)
```

This logic is compliant with [Semantic conventions 1.35.0 / Non-normative / K8s attributes
Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/), it's what we want.

cyrille-leclerc (approved):

cyrille-leclerc:
Should we review our logic to set `service_instance_id` and ensure that it also complies with [Semantic conventions 1.35.0 / Non-normative / K8s attributes - Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/)?

cyrille-leclerc:
Do we have a documentation PR to explain this evolution?

petewall:
> Should we review our logic to set `service_instance_id` and ensure that it also complies with [Semantic conventions 1.35.0 / Non-normative / K8s attributes - Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/)?

Good call. We can set that, even if we don't pass it through by default. It would be a highly cardinal label, since it includes pod name. Perhaps its set as structured metadata?

cyrille-leclerc:
> > Should we review our logic to set `service_instance_id` and ensure that it also complies with [Semantic conventions 1.35.0 / Non-normative / K8s attributes - Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/)?
> 
> Good call. We can set that, even if we don't pass it through by default. It would be a highly cardinal label, since it includes pod name. Perhaps its set as structured metadata?

We don't store it as a structured metadata?
IMO his is yet another reason to embrace OTel pod logs making `gathermedhod: filelog` the default.

PR #1740, opened 2025-07-08
feat: promql tenant id

MatthiasDrewsCS:
With `promql` queries it is actually possible to query metrics from Mimir instead of Prometheus. When using multi tenancy an additional header `X-Scope-OrgID` is required to retrieve specific metrics.

This PR introduces a new optional environment variable to make this possible.


petewall:
This is really great, but in short time, I'm going to be migrating the test infrastructure from this got repository to the new grafana/helm-chart-toolbox. Can you make the same change in the query-test there?

MatthiasDrewsCS:
Thanks for the hint. Migrated the code and created a new PR: https://github.com/grafana/helm-chart-toolbox/pull/8

petewall (approved):

PR #1772, opened 2025-07-16
Always ensure __tenant_id__ is kept

petewall:


hotsezus:
@petewall Hi! It looks like you missed `labelsToKeep` in [Node Logs](https://github.com/grafana/k8s-monitoring-helm/blob/aecf0e6b0521680258f4cf7182afc2b037ebea32/charts/k8s-monitoring/charts/feature-node-logs/values.yaml#L45) and [Cluster Events](https://github.com/grafana/k8s-monitoring-helm/blob/aecf0e6b0521680258f4cf7182afc2b037ebea32/charts/k8s-monitoring/charts/feature-cluster-events/values.yaml#L43) feature.
Even though these are node logs and we cant route them to some tenant based on kubernetes namespace label, we can still route them to some static tenant
```hcl
stage.tenant {
   value = "kube-audit"
}
```
So its not some kubernetes-pods-only thing to keep `__tenant_id__` label - its common for every logs processing config

petewall:
@hotsezus fixing that here: https://github.com/grafana/k8s-monitoring-helm/pull/1780

PR #1777, opened 2025-07-17
Add the ability to set server and enforcement parameters

petewall:


dehaansa (approved):
LGTM. Only thought is a question, whether you want the default values to be written literally vs allowing the defaults in Alloy/OTel to take affect when no values are given. I'm not sure what the standard patterns are for that in the k8s-monitoring chart.

petewall:
Updated to fix the issues as well as use alloy's defaults.

PR #1778, opened 2025-07-18
expose aggregation_cardinality_limit on spanmetrics

### LINKED ISSUE #1771: Support aggregation_cardinality_limit on spanmetrics
`dimensions_cache_size` is deprecated according to the Alloy [docs](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.connector.spanmetrics/#arguments) in favor of `aggregation_cardinality_limit`, which is much better anyway.

The field is not yet exposed on the Helm chart.

tarek-sater-pf:
Fix #1771 

Expose `aggregation_cardinality_limit` for spanmetrics. We still have to mark `dimensions_cache_size` as deprecated.

Also not sure if `1000` is a sane default value.


petewall (approved):
Awesome!
Could you add a deprecation note for the dimensionsCacheSize field in the values.yaml?

```
    # -- How many dimensions to cache. DEPRECATED, please use aggregationCardinalityLimit instead.
```

tarek-sater-pf:
I think we still need to add logic not to set `dimensions_cache_size` if `aggregationCardinalityLimit` is specified.

Edit: I tested it with both set. Didn't find any errors, and it seems like it takes the aggregationCardinalityLimit only. Would be great if someone more knowledgeable can confirm.

Otherwise this PR is ready to be merged

PR #1781, opened 2025-07-18
Add a profiles receiver feature

petewall:
Also create docs and files to make creating new features easier


petewall:
+ @simonswine because it has the word "profiles" in it 😉

rlankfo (approved):

PR #1800, opened 2025-07-23
Allow passing destinations as a map

thandleman-r7:
This change allows passing a map of destinations in addition to the existing array. The key in each map will be used as the name.

## Changes
- Adds `destinationsMap` value 
- Adds a new template `zzz-destinations.yaml` that dynamically parses the map, and updates the values in `.Values.destinations` with the new destinations. Helm parses templates in reverse order, so the destinations value will be updated before validations or any other templates are parsed.
- Updated generated destinations definitions for `values.schema.json` 

## Motivation
We are using this chart to install the monitoring stack via Argo to hundreds of clusters. We have a set of default values that everyone will need, and an optional values file for any cluster-specific overrides that are necessary. This is a relatively common pattern.

Because destinations is a list, if we want to override just one key-value for a single destination, we have to overwrite the entire list.

By adding `destinationsMap`, we make it easier for GitOps-based workflows to configure monitoring.

### DIFF
--- a/charts/k8s-monitoring/README.md
+++ b/charts/k8s-monitoring/README.md
@@ -119,6 +119,30 @@ destinations:
       tenantIdFrom: env("LOKI_TENANT_ID")
 ```
 
+Alternatively, you may also define a map of destinations. The key for each destination in the map will be used as the name.
+
+```yaml
+destinationsMap:
+  hostedMetrics:
+    type: prometheus
+    url: https://prometheus.example.com/api/prom/push
+    auth:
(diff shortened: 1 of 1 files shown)


jwhitcraft:
@petewall what are you thoughts on this?

petewall:
> @petewall what are you thoughts on this?

I'll be tackling this soon. Things got side tracked with bugfixes that need to come in to a 3.2.3 release. I'm still planning on adopting this for 3.3.

PR #1830, opened 2025-08-05
fix(crd-validation): 🐛 remove bad double quote in crd validation

jmapro:
Removes the empty double quote that breaks the Helm chart when `alloy-operator.deploy` is set to `false` to avoid error

```
 executing "crdValidation" at <append>: wrong number of args for append: want 2 got 3
```

### DIFF
--- a/charts/k8s-monitoring/templates/_crd-validation.tpl
+++ b/charts/k8s-monitoring/templates/_crd-validation.tpl
@@ -2,7 +2,7 @@
 {{- $crdURL := printf "https://github.com/grafana/alloy-operator/releases/download/alloy-operator-%s/collectors.grafana.com_alloy.yaml" (index .Subcharts "alloy-operator").Chart.Version }}
 {{- if .Release.IsInstall }}
   {{- if not (.Capabilities.APIVersions.Has "collectors.grafana.com/v1alpha1/Alloy") }}
-    {{- if not (index .Values "alloy-operator").crds.deployAlloyCRD }}
+    {{- if not (dig "alloy-operator" "crds" "deployAlloyCRD" true .Values.AsMap) }}
       {{- $msg := list "" (printf "The %s Helm chart v3.0 requires the Alloy CRD to be deployed." .Chart.Name) }}
       {{- $msg = append $msg "Please set:" }}
       {{- $msg = append $msg "alloy-operator:" }}
--- a/charts/k8s-monitoring/templates/_crd-validation.tpl
+++ b/charts/k8s-monitoring/templates/_crd-validation.tpl
@@ -8,7 +8,7 @@
       {{- $msg = append $msg "alloy-operator:" }}
       {{- $msg = append $msg "  crds:" }}
       {{- $msg = append $msg "    deployAlloyCRD: true" }}
-      {{- $msg = append $msg "" "Or install the Alloy CRD manually:" }}
+      {{- $msg = append $msg "Or install the Alloy CRD manually:" }}
       {{- $msg = append $msg (printf "kubectl apply -f %s" $crdURL) }}
       {{- fail (join "\n" $msg) }}
     {{- end }}


petewall:
The double-quote is not the issue. What that's doing is appending a blank string along with the string with the message. Then, later, when we run `(join "\n" $msg)` it turns the blank string into an extra newline.

jmapro:
my bad, the issue I linked does fully correspond to my issue.
I have the following error when I deploy. The double quote is an argument and the message too and it breaks the append function

```
 executing "crdValidation" at <append>: wrong number of args for append: want 2 got 3
``` 

I've updated the issue description to reflect this more accurately. :)

petewall:
The actual issue appears to be this line:
```
    {{- if not (index .Values "alloy-operator").crds.deployAlloyCRD }}
```

If the alloy operator is disabled, it'll likely not load the `crds` object.

We should update the logic to be:
```
  // Check if the Alloy CRD exists
  {{- if not (.Capabilities.APIVersions.Has "collectors.grafana.com/v1alpha1/Alloy") }}
    // Check if Alloy CRD will be deployed. This needs to be updated
    {{- if not (index .Values "alloy-operator").crds.deployAlloyCRD }}
...
      {{- fail (join "\n" $msg) }}
    {{- end }}
  {{- end }}
{{- end }}
```

I bet it needs to be split into two checks:

1. Is alloy-operator disabled? If not, it should print a message saying "alloy operator is required for version 3. please enable it"
2. else, is alloy-operator enabled, but deploy alloy crds is disabled?

The second check probably could be updated to:
```
{{- if not (dig "alloy-operator" "crds" "deployAlloyCRD" true .Values) }}
```

which would be more tolerant.

jmapro:
I've updated the logic with our recommendation which fixes https://github.com/grafana/k8s-monitoring-helm/issues/1707 . I used `.AsMap` due to this issue http://github.com/helm/helm/issues/9266.

I'm still convinced that the following code can't work due to the empty double quotes
```
      {{- $msg = append $msg "" "Or install the Alloy CRD manually:" }}
```

jmapro:
> Is alloy-operator disabled? If not, it should print a message saying "alloy operator is required for version 3. please enable it"

I don't think we should check for alloy-operator here. If the deployment of the operator is managed outside this chart it will not work.

For exemple, the operator is deployed with an account with cluster wide privileges and this chart deployed with an account with fewer privileges. In that case, we disable alloy-operator here.

charts/k8s-monitoring/templates/_crd-validation.tpl
```diff
@@ -2,13 +2,13 @@
...
   {{- if not (.Capabilities.APIVersions.Has "collectors.grafana.com/v1alpha1/Alloy") }}
-    {{- if not (index .Values "alloy-operator").crds.deployAlloyCRD }}
+    {{- if not (dig "alloy-operator" "crds" "deployAlloyCRD" true .Values.AsMap) }}
       {{- $msg := list "" (printf "The %s Helm chart v3.0 requires the Alloy CRD to be deployed." .Chart.Name) }}
       {{- $msg = append $msg "Please set:" }}
       {{- $msg = append $msg "alloy-operator:" }}
       {{- $msg = append $msg "  crds:" }}
       {{- $msg = append $msg "    deployAlloyCRD: true" }}
-      {{- $msg = append $msg "" "Or install the Alloy CRD manually:" }}
+      {{- $msg = append $msg "Or install the Alloy CRD manually:" }}
```
petewall, line 11:
```suggestion
      {{- $msg = append $msg "" }}
      {{- $msg = append $msg "Or install the Alloy CRD manually:" }}
```

petewall: My bad, append totally only takes one extra argument. For the sake of readability, can we add the newline back with this change?

petewall (approved):

PR #1864, opened 2025-08-11
Add annotation control for profiles to v1.

petewall:
Should not change the behavior, because it's still opt-out, but can use annotations to disable scrape.


petewall:
Mostly just copy-paste from the v3 change I made a while back

PR #1879, opened 2025-08-13
Allow declaring scrape protocols and scrape classic histograms

petewall:
Only set currently via the `global` values section.

Also remove the last vestiges of alloyModules


petewall:
cc @SeamusGrafana

PR #1880, opened 2025-08-14
featx(spanmetrics): add transform support for spanmetrics

mbaykara:


mbaykara:
I made the changes according to your suggestion.

PR #1888, opened 2025-08-15
Force hooks on to linux nodes

petewall:


petewall:
Found while running the `eks-with-windows` platform test.

PR #1896, opened 2025-08-19
Fix Typo in Values Setting Name: Correct `scrapeProcotols` to `scrapeProtocols`

SeamusGrafana:
Fix Typo in Values Setting Name: Correct `scrapeProcotols` to `scrapeProtocols`

This was recently added in v3.3 and #1879


petewall (approved):
🤦🏻‍♂️ 
thanks for finding and fixing that!

PR #1917, opened 2025-08-23
feature(sys-env): Adding sys.env functions to remoteConfig collector

AzgadAGZ:
This pull request enhances the remote configuration flexibility for Alloy collectors in the Kubernetes monitoring Helm chart. The main improvement is the addition of a new field that allows users to specify the remote configuration URL using raw Alloy references, making it easier to load values dynamically (e.g., from environment variables).

**Remote configuration improvements:**

* Added a new `urlFrom` field in `alloy-values.yaml` to allow specifying the remote config URL using raw Alloy references, such as environment variables or config maps.
* Updated the `_collector_remoteConfig.tpl` template to prioritize `urlFrom` if set, falling back to the standard `url` field otherwise.


AzgadAGZ:
@petewall another quick one, basically what I faced was an issue to pass that value from a secret, since I want to avoid anything exposed in the values.yaml, even the URLs

petewall:
This looks good! Can you regenerate schema files and add a line in the unreleased section of the CHANGELOG?

AzgadAGZ:
> This looks good! Can you regenerate schema files and add a line in the unreleased section of the CHANGELOG?

Solved Chief! 🫡

PR #1938, opened 2025-08-28
Add Database Observability

petewall:
This enhances the existing MySQL integration to add the database observability components. I opted to use the existing `integrations` feature, rather than creating a new `databaseObservability` feature, since it'll be much easier for existing MySQL integration users to update to use DBO11y.

This only adds MySQL for now, but I can add PostgreSQL soon in a future release.

The change to [docs/examples/features/database-observability/mysql/alloy-singleton.alloy](https://github.com/grafana/k8s-monitoring-helm/pull/1938/files#diff-e01439faacb8956274d79e3ab7c325e46d378679635523f16885a06071e0ed8e) has the easiest to review Alloy config.

I am testing this with integration and platform tests. The platform test, which uses Grafana Cloud, does indeed populate dbo11y.

PR #1990, opened 2025-09-10
Update the way service_name, service_namespace, and service_instance_id are calculated for profiling

### LINKED ISSUE #1678: Update profiling rules to extract `service_name` and `namespace`
Put this into the default config:
```
profiling:
      enabled: true
      java:
        enabled: true
        annotationSelectors:
          profiles.grafana.com/inject-java: "true"
        extraDiscoveryRules: |
          rule {
            source_labels = ["__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name"]
            regex = "(.+)"
            target_label = "service_name"
          }
        
          rule {
            action="replace"    
            regex = "(.+)"
            source_labels = ["__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_namespace"]
            target_label = "namespace"
          }
```

petewall:
Fixes #1938
Fixes #1678


rlankfo (approved):

cyrille-leclerc:
@petewall please review my update of the PR description. Is it correct that this PR setts labels on profiles?
If so, i'm sorry it's outside my living memory but I have 2 questions:
* Shouldn't we update `namespace` with the value of `service_namespace` to correlate with Grafana App O11y? Note that if we overwrite `namespace` with the value of `service_namespace`, I'm wondering if we could also set `k8s_namespace_name` to keep the information. I think that @MattiasSegerdahl's use case needs `namespace` to be set with the value of `service_namespace`
* Is Pyroscope expecting `service_instance_id` to be a label? I thought we had another metadata mechanism in Pyroscope to convey this information?

@simonswine Please check me

simonswine:
> * Shouldn't we update `namespace` with the value of `service_namespace` to correlate with Grafana App O11y? Note that if we overwrite `namespace` with the value of `service_namespace`, I'm wondering if we could also set `k8s_namespace_name` to keep the information. I think that @MattiasSegerdahl's use case needs `namespace` to be set with the value of `service_namespace`

I guess that would be another breaking change (other than `service_name`), so the right time to this would be now. I think we should always set the `k8s_namespace_name`, no matter if we decide to do the other change and if we do it we should similarily provide `extraDiscoveryRules` to retain the old behaviour.

> * Is Pyroscope expecting `service_instance_id` to be a label? I thought we had another metadata mechanism in Pyroscope to convey this information?

No that's not used in any special way, Pyroscope treats all labels the same once you select a "service_name". I guess typically `pod` and `container` are used.

petewall:
I think rewriting `namespace` to match `service_namespace` as well as adding `k8s.namespace.name` does two things that make me uncomfortable:
* It potentially breaks `namespace`, which was set before.
* It sets an odd precedent. If we're setting `k8s.namespace.name`, do we also add `k8s.pod.name` and `k8s.container.name`? Do we keep or just drop `pod` and `container`?

Following both of these paths definitely start to feel too much like breaking changes as well as a large increase in cardinality.

petewall:
Should I not add setting `service_instance_id` in this change?

PR #1994, opened 2025-09-11
Pdc-agent helm chart feature

peterolivo:
This is my second take at getting the pdc-agent into the helm chart.
I ran into some issues with wrapping the chart in a feature without referencing the dependent chart values. I would take some advice there.
Also the readme and metrics config seems a bit bulky, it was generated. They can be edited if need be.


petewall:
Ehh... I think I'm going to close this.
The PDC agent is great for connecting datasources inside the cluster, but this feature really isn't related to the monitoring information within the cluster. As in, if we included it, it would just be a toggle to add a deployment, but wouldn't likely affect any other aspect of the Helm chart.

PR #2083, opened 2025-10-17
Ensure that service graph generator utilizes destination defaults

petewall:
Without this, the default values in the OTLP and Prometheus destinations, like:
```
processors:
  batch:
    enabled: true
```

Was undefined. So we ran into null issues unless the default was re-defined.


charts/k8s-monitoring/templates/alloy-servicegraph.yaml
```diff
@@ -14,32 +14,34 @@
 
     {{- /* Destination components*/}}
     {{- $metricsDestinationNames := include "destinations.get" (dict "destinations" $.Values.destinations "type" "metrics" "ecosystem" "otlp" "filter" $destination.processors.serviceGraphMetrics.destinations) | fromYamlArray -}}
-    {{- $destinations := list }}
```
petewall: This variable was just unused

rlankfo (approved):

PR #2084, opened 2025-10-19
chore(auto-instrumentation): add verbosity when app O11y not enabled

mbaykara:
When users enable Beyla auto-instrumentation without enabling Application Observability, only metrics are collected through the cluster-metrics feature.
In this setup, traces are not forwarded.
This change adds a warning message to users when auto-instrumentation is enabled but Application O11y is not, helping them understand why traces may be missing.

### DIFF
--- a/charts/k8s-monitoring/charts/feature-application-observability/templates/_notes.tpl
+++ b/charts/k8s-monitoring/charts/feature-application-observability/templates/_notes.tpl
@@ -15,26 +15,26 @@ Gather application data via {{ include "english_list" $receivers }} {{ $receiver
 
 {{- define "feature.applicationObservability.notes.actions" }}
 Configure your applications to send telemetry data to:
-{{- if .Values.receivers.otlp.grpc.enabled }}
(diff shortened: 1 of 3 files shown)


petewall:
This is a great addition. I'm wondering if the logic and the wording of the note belongs in `charts/k8s-monitoring/charts/feature-auto-instrumentation/templates/_notes.tpl`. I recognize that we don't currently pass in all values:
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/templates/NOTES.txt#L29

but we could update that.

Maybe something like this:
```
  {{ include (printf "feature.%s.notes.actions" $feature) (dict "Values" (index $.Values $feature) "ParentValues" $.Values "Collector" $Collector) | indent 0 }}
```

mbaykara:
Awesome, I'll make the changes, was at PromCon

mbaykara:
Hey @petewall, I changed as as you said, moving into respective chart.

petewall (approved):

PR #2202, opened 2025-12-03
Fix minor typos in k8s-monitoring chart's README

mihaelmiklec:
Caught these two while glancing the README, not much else to say. :)


petewall:
Ooh, good catch! Thanks for that!

Issue #1946, opened 2025-09-02
Add support for include/exclude block in otelcol.processor.attributes


cristian-1:
It is currently not possible to configure include and exclude blocks (https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.processor.attributes/#include-block) in the Attributes processor (https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/docs/destinations/otlp.md#attributes-processor)

This would be helpful if we want to target only a subset of data

cristian-1:
Is there any plan for this? It's the only thing keeping us from using app observability instead of a full extraConfig.

This was possible with chart v1: https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring-v1/README.md#otel-receivers-processors

This is how is configured in Alloy custom config:

```
    otelcol.processor.attributes "otlp_endpoint" {
      include {
        match_type = "strict"
        services   = ["api-gateway", "${GRAFANA_ALLOY_CLUSTER}-intranet", "${GRAFANA_ALLOY_CLUSTER}-internet"]
      }
      action {
        key = "http.url"
        action = "extract"
        pattern = "(?:https?:\\/\\/)?(?P<deployment>[a-z][\\w-]+)\\.(?P<namespace>[a-z][\\w-]+)(?:\\..+){2,}\\/.*"
      }
      action {
        key = "http.target"
        action = "extract"
        pattern = "(?:https?:\\/\\/)?\\/(?P<namespace>[a-z][\\w-]+)\\/(?P<deployment>[a-z][\\w-]+)\\/.*"
      }
      output {
        traces = [otelcol.processor.transform.otlp_endpoint.input]
      }
    }
```

 Thanks

PR #2209, opened 2025-12-10
feat(otelcol.processor.attribute): support include/exclude blocks

mbaykara:


mbaykara:
support #1946

petewall:
Were there changes to `charts/k8s-monitoring/destinations/otlp-values.yaml`? I don't see it in the diffs.

mbaykara:
> Were there changes to `charts/k8s-monitoring/destinations/otlp-values.yaml`? I don't see it in the diffs.

I am fixing with some other changes

mbaykara:
We have to make here a design decision,
if we want to support attribute based processing which is limited to `traces` and `logs` but metrics are only supported by `metricNames`, user will be only able output traces and logs. 

> For metrics, metric_names must be specified with a valid non-empty value for a valid configuration. The span_names, span_kinds, log_bodies, log_severity_texts, log_severity, services, [attribute](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.attributes/#attribute), [resource](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.attributes/#resource), and [library](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.attributes/#library) attributes are invalid.

If we support resource attribute level, this will/can be used for all signals.
e.g
```yaml
    processors:
      attributes:
        include:
          matchType: "strict"
          resources:
            - key: "service.name"
              value: "api-gateway"
            - key: "service.name"
              value: "payment-service"
            - key: "service.name"
              value: "user-service"
        actions:
          - key: "http.url"
            action: "extract"
            pattern: '(?:https?://)?(?P<deployment>[a-z][\w-]+)\.(?P<namespace>[a-z][\w-]+)(?:\..+){2,}/.*'
```

Imho supporting resource attribute level is most flexible way, user is still able to promote an attribute to resource attribute and then process it as above. Wdyt? @petewall and @rlankfo

rlankfo:
> If we support resource attribute level, this will/can be used for all signals.

Ideally we could support attributes as well. If the user adds an attribute config for metric signals, is this ignored or does it cause the collector to fail?

Perhaps we could have some validations or warnings that specify which match types won't apply to specific signals if the user has it configured?

mbaykara:
> > If we support resource attribute level, this will/can be used for all signals.
> 
> Ideally we could support attributes as well. If the user adds an attribute config for metric signals, is this ignored or does it cause the collector to fail?
> 
> Perhaps we could have some validations or warnings that specify which match types won't apply to specific signals if the user has it configured?

When metrics are enabled, attributes (along with services and libraries) causes the collector to fail,  it's not ignored. The Alloy error is: "services", "attributes" and "libraries" are not valid for metrics.

We can add support for resources for all signals and support signal specific support which will require disable i.e metrics for the destination. 

Another option will be separate attribute processors for different signal types if you need different filtering per type.

I've added validation that will fail early with clear error messages explaining which fields work for which signals, so users get helpful feedback before deployment.

Issue #2237, opened 2026-01-08
node_exporter already excludes erofs, but k8s-monitoring still passes fs-types-exclude without erofs


tyuchx:
We recently installed k8s-monitoring and noticed a mismatch between the chart defaults and the upstream node_exporter behavior related to filesystem metrics collection.

Starting with **node_exporter v1.10.0**, `filesystem_linux.go` excludes `erofs` internally. However, when installing **k8s-monitoring chart v3.7.1**, the generated `node_exporter` DaemonSet still passes an explicit `--collector.filesystem.fs-types-exclude` flag that does **not include `erofs`**:

```
--collector.filesystem.fs-types-exclude=^(autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|ramfs|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tmpfs|tracefs)$
```

Updating the default filesystem exclusion list to include `erofs` would help keep k8s-monitoring aligned with upstream behavior.

### Chart / version details

- k8s-monitoring chart version: **3.7.1**
- Component: `clusterMetrics.node-exporter`

### References

- node_exporter v1.10.0 filesystem changes:  
  https://github.com/prometheus/node_exporter/blob/v1.10.0/collector/filesystem_linux.go

PR #2257, opened 2026-01-18
chore(cluster-metrics): add erofs to node-exporter filesystem exclusions

tyuchx:
Add erofs (Enhanced Read-Only File System) to the filesystem types excluded by node-exporter. This prevents unnecessary metric collection from read-only erofs filesystems commonly used in container images and embedded systems.

Ref #2237


tyuchx:
Hi @petewall ,

I ran make build, but there were no changes to commit. Could you please provide more detailed steps or clarify if I missed anything?

Issue #100, opened 2023-08-30
Discussion: Should we set the default logging format to CRI?


petewall:
The [default](https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/values.yaml#L334-L337) logging format is currently set to docker. Kubernetes deprecated the docker runtime in 1.20 in favor of CRI.

Should this stay or should the default be migrated to CRI?

rgeyer:
So... Interestingly, the container runtime is exposed in the discovery labels, which I just learned.

Further, there is an agent module which handles a lot of this magic, including fetching, and passing along the container runtime to inform the processing stage. 🤔 

https://github.com/grafana/agent-modules/blob/main/modules/kubernetes/logs/targets/logs-from-worker.river#L54-L72

petewall:
Ooh, I really like this approach.

PR #2282, opened 2026-01-29
fix: Set CRI as default logs processor if runtime is unset

aleksanderaleksic:
# Problem
#127 introduced logic for selecting the correct runtime pipeline.
We have noticed that on pod startup and for the first second the logs is prefixed with `stdout F` but after that the prefixing is gone. 

In the image below you can see that the `stdout F` logs correlate almost always with pod startup. At 9:30 I forced alloy to use CRI and the problem disappeared. 
Screenshot 2026-01-29 at 10 48 45

I think this is because the `tmp_container_runtime` label is not always available on startup after pod discovery leading to that no logs pipeline is selected by alloy. 

We are just using the default configuration for the `3.7.1` release.
```yaml
podLogs:
  enabled: true
  destinations:
    - grafana-cloud-otel
```
----
# Suggested solution

This PR sets CRI (based on the discussion in #100) as the default  pipeline if `tmp_container_runtime` is not set.

### DIFF
--- a/charts/k8s-monitoring/charts/feature-pod-logs-via-kubernetes-api/templates/_common_log_processing.alloy.tpl
+++ b/charts/k8s-monitoring/charts/feature-pod-logs-via-kubernetes-api/templates/_common_log_processing.alloy.tpl
@@ -38,25 +38,6 @@ loki.process "pod_logs" {
     }
   }
 
-  // The default processing stage if tmp_container_runtime is not set or empty.
-  stage.match {
-    selector = "{tmp_container_runtime=""}"
-    // the docker processing stage extracts the following k/v pairs: log, stream, time
-    stage.cri {
-{{- if .Values.cri.maxPartialLines }}
-      max_partial_lines = {{ .Values.cri.maxPartialLines }}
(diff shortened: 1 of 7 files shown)


aleksanderaleksic:
I guess a alternative solution could be to filter out the pods in the discovery where the `tmp_container_runtime=""`, but its probably still a good idea to have a fallback processor?

petewall:
I like the concept of this PR, but I might riff on it a bit. Make it configurable, so people can choose to use this or not.
We also need to regenerate the generated files, but I can take care of that.

aleksanderaleksic:
@petewall  I like your approach better 👍

As a temporary workaround I did add a `extraDiscoveryRule`:
```yaml
extraDiscoveryRules: |
  rule {
    action = "drop"
    source_labels = [ "tmp_container_runtime" ]
    regex = ""
  }
```
"delaying" the logs ingesting but selecting the correct processor :)

defanator:
@petewall would it be possible to backport this to GKM v1?

(https://github.com/grafana/alloy/issues/5319)

PR #2294, opened 2026-02-02
Add support for write-ahead-log (wal) setting for loki destination

nooboo:
In our workflow we want to enable the wal (write-ahead-log) in loki.wrote component. 
With this small improvement you can enable it in the loki destination by proposed values changes.

```yaml
writeAheadLog:
  enabled: false
  drainTimeout: 30s
  minReadFrequency: 250ms
  maxReadFrequency: 1s
  maxSegmentAge: 1h
```


petewall:
Thanks for sending this PR! It needed some updating, so I took it over and put it in a new PR. I did credit you in the Changelog, though!

PR #2310, opened 2026-02-12
feat: enrich pod logs with namespace and pod labels

mbaykara:
Similar to metricEnrichment feature, following the same pattern using [loki.enrich](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.enrich/)
allows to add dynamic labels without changing the gatherMethod


petewall:
This is now possible with the pod logs feature themselves, or with the k8sEnrichment data processor.

PR #2352, opened 2026-03-04
fix(servicegraph): preserve user extraEnv in service graph collector

rafix:
## Problem

The service graph template (`alloy-servicegraph.yaml`, line 11) uses Helm's `merge` to build collector values:

```go
$collectorValues := merge (dict "fullnameOverride" $collectorName "alloy" (dict "extraEnv" $extraEnvContent)) ($destination.processors.serviceGraphMetrics.collector | default dict)
```

Helm's `merge` gives priority to the **first** dict. Since the template hard-codes `alloy.extraEnv` with just `POD_NAME`, any user-provided `extraEnv` entries via `collector.alloy.extraEnv` are silently dropped.

This means users cannot inject custom environment variables into the service graph Alloy pod — which breaks auth when the Prometheus destination uses `passwordFrom: env("...")` since the env var doesn't exist in the service graph pod.

## Fix

1. **`concat`** the required `POD_NAME` env var with user-provided `extraEnv` entries (so both are preserved)
2. Switch from **`merge`** to **`mergeOverwrite`** so user collector values take priority over template defaults

```go
$userExtraEnv := dig "alloy" "extraEnv" list ($destination.processors.serviceGraphMetrics.collector | default dict)
$combinedExtraEnv := concat $extraEnvContent $userExtraEnv
$collectorValues := mergeOverwrite (dict "fullnameOverride" $collectorName "alloy" (dict "extraEnv" $combinedExtraEnv)) ($destination.processors.serviceGraphMetrics.collector | default dict)
```

## Reproduction

Configure a service graph with a custom env var:

```yaml
destinations:
  - name: my-otlp
    type: otlp
    auth:
      type: basic
      username: "12345"
      passwordFrom: env("MY_API_KEY")
    processors:
      serviceGraphMetrics:
        enabled: true
        collector:
          alloy:
            extraEnv:
              - name: MY_API_KEY
                valueFrom:
                  secretKeyRef:
                    name: my-secret
                    key: api-key
```

**Before fix:** The `MY_API_KEY` env var is dropped; only `POD_NAME` exists in the pod. Auth fails with 401.  
**After fix:** Both `POD_NAME` and `MY_API_KEY` are present. Auth works correctly.

### DIFF
--- a/charts/k8s-monitoring/templates/alloy-servicegraph.yaml
+++ b/charts/k8s-monitoring/templates/alloy-servicegraph.yaml
@@ -7,10 +7,9 @@
 
     {{- $maxLength := 51 }}{{/* This limit is from the `controller-revision-hash` pod label value*/}}
     {{- $collectorName := printf "%s-%s" $.Release.Name (include "helper.k8s_name" (printf "%s-servicegraph" $destination.name)) | trunc $maxLength | trimSuffix "-" | lower }}
-    {{- $extraEnvContent := list (dict "name" "POD_NAME" "valueFrom" (dict "fieldRef" (dict "apiVersion" "v1" "fieldPath" "metadata.name"))) }}
-    {{- $userExtraEnv := dig "alloy" "extraEnv" list ($destination.processors.serviceGraphMetrics.collector | default dict) }}
-    {{- $combinedExtraEnv := concat $extraEnvContent $userExtraEnv }}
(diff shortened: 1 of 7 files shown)


petewall:
Thanks for the PR, @rafix ! I'll take a look

petewall:
I kind of want to introduce a helper to do this. This isn't the only place where we merge Env vars. Let me update the PR...

petewall (approved):

PR #2507, opened 2026-04-17
Restricting default privileges for Alloy

petewall:


petewall:
Ehh, this might need to go to 5.0?
It could be a breaking change.

PR #2539, opened 2026-04-27
Use external_labels instead of write_relabel_config for cluster labels

petewall:
cc @kgeckhart


kgeckhart (approved):
ty for the fast turn around!

PR #2553, opened 2026-05-02
Add Argo CD external-secrets and ApplicationSet examples

jmichalek132:
Adds two new Argo CD deployment-alternative manifests alongside the existing `k8s-monitoring-application.yaml`, so users who already use External Secrets Operator, Sealed Secrets, SOPS, Vault, or similar have a complete starting point that doesn't put credentials in the Application manifest, and users running Argo CD across many clusters have an `ApplicationSet` example with the cluster generator.

The new manifests intentionally mirror the existing example's full feature set (clusterMetrics, costMetrics, hostMetrics with Linux/Windows/energy, clusterEvents, podLogsViaLoki, full telemetryServices including OpenCost) so the only diff between them and the existing example is credential sourcing. Validated locally via `helm template`: both new manifests produce a 2952-line render byte-identical to the existing example modulo `cluster.name`. The README cross-links to the [Grafana Cloud Argo CD configuration page](https://grafana.com/docs/grafana-cloud/monitor-infrastructure/kubernetes-monitoring/configuration/config-other-methods/argocd-config/) that walks users through these manifests step by step; that docs page is being updated in parallel to match the new shape.

The `ApplicationSet` uses `{{ .nameNormalized }}` for `metadata.name` so cluster names containing underscores or uppercase characters don't produce DNS-1123-invalid Application names, with a README caveat that `nameNormalized` is not injective (`prod_us` and `prod-us` both normalize to `prod-us`) and users should ensure their cluster names are unique after normalization.

Question for maintainers: the existing `argocd/test/` rig only covers `k8s-monitoring-application.yaml`. Worth extending it (or adding a sibling `test/` directory) to exercise the two new manifests in CI? The external-secrets variant would need its `Secret` applied as a test dependency; the `ApplicationSet` variant would need a labelled Cluster secret on the kind node so the generator has something to iterate over. Happy to add either as a follow-up commit on this PR if you want it before merge.

### DIFF
--- a/charts/k8s-monitoring/docs/examples/deployment-alternatives/argocd/README.md
+++ b/charts/k8s-monitoring/docs/examples/deployment-alternatives/argocd/README.md
@@ -20,7 +20,10 @@ Three variants are provided:
 - [`k8s-monitoring-applicationset.yaml`](./k8s-monitoring-applicationset.yaml): an `ApplicationSet` using the
   [cluster generator](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Cluster/) to
   fan the same configuration out to every Argo CD cluster carrying the `kubernetes-monitoring: enabled` label.
-  Combine with the external-secrets variant when each managed cluster has its own credentials Secret.
+  Combine with the external-secrets variant when each managed cluster has its own credentials Secret. The template
+  uses `{{ .nameNormalized }}` for the generated `Application.metadata.name`; ensure your Argo CD cluster names are
+  unique after normalization (lowercase + non-DNS characters replaced with `-`), since names like `prod_us` and
+  `prod-us` both normalize to `prod-us` and would generate colliding Applications.
 
 ## Setting `null` values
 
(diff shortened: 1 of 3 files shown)


jmichalek132:
The failing `Integration Test (docs/examples/deployment-alternatives/argocd/test)` looks like a pre-existing flake. This PR does not modify `k8s-monitoring-application.yaml` (the manifest that test exercises) or any code path that affects its rendering — the two new YAML files are siblings in the same directory and aren't covered by `argocd/test/test-plan.yaml`.

The same job failed on `main` in run [25081561097](https://github.com/grafana/k8s-monitoring-helm/actions/runs/25081561097) about four hours before mine, with the same symptom (`context deadline exceeded`, `grafana_kubernetes_monitoring_build_info{cluster="argocd-deployment-test"}` returning empty 10 consecutive times). The chart appears to deploy but metrics don't reach the in-cluster Prometheus before the test timeout.

Could you re-run the failed job? I don't have rerun permission from the fork.

petewall (approved):

PR #2596, opened 2026-05-12
fix(spanmetrics): quote exclude_dimensions in rendered config

savannahostrowski:
Apologies if I missed some repo conventions here and should have opened an issue first! Seemed like a simple enough fix to just put up the PR 🙂 

Right now, setting `connectors.spanMetrics.excludeDimensions` to a non-empty list causes Alloy to fail to parse the rendered config:

```
Error: component "status.code" does not exist or is out of scope
```

The values are rendered as unquoted bareword identifiers; Alloy interprets dotted names like `status.code` as component references instead of string literals.

Seems like we just need to serialize the list properly via `| toJson`. I've added that as well as a test to exercise the path.

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+*   Fix `spanMetrics.excludeDimensions` rendering: list values are now properly quoted as strings instead of unquoted bareword identifiers, which caused Alloy to fail to parse the rendered config. (#2596) (@savannahostrowski)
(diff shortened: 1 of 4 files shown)


petewall (dismissed):

petewall:
@savannahostrowski, thanks for the PR! Often an issue is good form, but I wont let that stop me from accepting a fix! Haha.

One last request, can you update the charts/k8s-monitoring/CHANGELOG.md file in a new `## Unreleased` section with a sentence of your fix? Then I'll get it merged and in the next patch release.

petewall (approved):

Issue #2582, opened 2026-05-09
Chart 4.1.x scrapeNativeHistograms: true crashing alloy-metrics pods


ceastman-r7:
When I set the following in the values.yaml I get the following error log from the alloy-metrics statefulset:

`
global:
  scrapeNativeHistograms: true
`

error log:
`
ts=2026-05-09T05:42:18.107755595Z level=error source=/src/alloy/internal/runtime/internal/controller/loader.go:939 msg="failed to evaluate config" controller_path=/ controller_id=cluster_metrics.feature trace_id=f48456dc65f0971bcf062c71fd8646a6 node=prometheus.scrape.kubelet err="decoding configuration: scrape_native_histograms is set to true, but PrometheusProto is not in scrape_protocols"
`

petewall:
Working on a fix for this.
In the meantime, if you set `global.scrapeProtocols = ["PrometheusProto", "OpenMetricsText1.0.0", "OpenMetricsText0.0.1", "PrometheusText0.0.4"]` it should fix this crash.

petewall:
Should be fixed by https://github.com/grafana/k8s-monitoring-helm/pull/2589

PR #2601, opened 2026-05-14
feat(validations): fail loudly when stabilityLevel is too low for experimental Alloy components

MattiasSegerdahl:
## What does this PR do?

The chart writes Alloy config that requires `stabilityLevel: experimental` whenever either of these is set:

- `global.convertClassicHistogramsToNhcb: true` (renders `convert_classic_histograms_to_nhcb = true` on prometheus.scrape components)
- a prometheus destination on `remoteWriteProtocol: 2` (renders `protobuf_message = \"io.prometheus.write.v2.Request\"` on prometheus.remote_write)

Without `stabilityLevel: experimental` on the affected collector the Alloy operator schedules the collector and then it crash-loops at init:

```
building component: using remote write v2 (protobuf_message=io.prometheus.write.v2.Request)
with endpoint <url> requires setting the stability.level flag to experimental
```

or, for NHCB:

```
... convert_classic_histograms_to_nhcb = true ... requires setting the stability.level flag to experimental
```

This is the natural follow-on to #2586 (`convertClassicHistogramsToNhcb`) and #2582 (auto-prepend `PrometheusProto`): the chart now invites users down a config path it knows needs experimental, but doesn't surface that to them until the collector is in CrashLoopBackOff.

This PR adds a chart-level validation that runs at `helm template` time:

- walks `features.list.enabled`,
- for each feature looks up the assigned collector (`collectors.getCollectorForFeature`) and the destinations it forwards to (`features.<feature>.destinations`),
- and if any (feature, destination) pair would render an experimental-only Alloy component, fails the render with the exact `collectors.<name>.alloy.stabilityLevel: experimental` snippet to set, plus a list of which feature+destination combinations triggered the requirement.

## Example

Stock 4.1.x values from a Grafana Cloud onboarding wizard (omitting destination credentials) plus the two recommended knobs from #2586 / #2582 / chart docs:

```yaml
cluster:
  name: production
destinations:
  grafana-cloud-metrics:
    type: prometheus
    url: https://prometheus-prod-X-prod-us-central-0.grafana.net/api/prom/push
    remoteWriteProtocol: 2
global:
  convertClassicHistogramsToNhcb: true
clusterMetrics:
  enabled: true
collectors:
  alloy-metrics:
    presets: [clustered]
```

Before this PR: \`helm install\` succeeds, alloy-metrics-0 starts and immediately enters CrashLoopBackOff. \`kubectl logs\` shows the \`requires setting the stability.level flag to experimental\` line.

After this PR:

```text
Collector \"alloy-metrics\" has stabilityLevel=\"generally-available\" but the chart will render Alloy components that require stabilityLevel=experimental:
  - feature \"clusterMetrics\" forwards to destinations.grafana-cloud-metrics, which uses remoteWriteProtocol: 2 and renders \`protobuf_message = \"io.prometheus.write.v2.Request\"\` (Alloy exposes this only at stability.level=experimental)
  - feature \"clusterMetrics\" forwards to a prometheus destination (grafana-cloud-metrics) and global.convertClassicHistogramsToNhcb: true renders \`convert_classic_histograms_to_nhcb = true\` on its scrape components (Alloy exposes this only at stability.level=experimental)

Please set:
collectors:
  alloy-metrics:
    alloy:
      stabilityLevel: experimental
```

## Why per-feature scoping

Earlier in development I tried a chart-wide check (\"if NHCB or RWv2 is set anywhere, every enabled collector needs experimental\"). That broke the existing \`tests/platform/grafana-cloud/database-observability\` fixture, where alloy-logs is enabled alongside a Prometheus destination on RWv2 but alloy-logs only forwards to Loki and never renders \`prometheus.remote_write\`. Walking features and only flagging the collector(s) actually hosting an experimental component avoids the false positive — there's a dedicated test for that scenario.

## Tests

\`charts/k8s-monitoring/tests/collector_stability_level_test.yaml\` adds five cases:

1. \`global.convertClassicHistogramsToNhcb: true\` with no override → fails with the right message
2. Prometheus destination on \`remoteWriteProtocol: 2\` with no override → fails with the right message
3. Both of the above + \`collectors.alloy-metrics.alloy.stabilityLevel: experimental\` → passes
4. Mixed Loki+Prometheus install with NHCB and RWv2, alloy-logs without experimental, alloy-metrics with experimental → passes (no false positive on the logs collector)
5. Neither flag set → passes

Full chart suite (\`make build && make test ALLOY_OPERATOR_VERSION=...\`): 175/175 unit tests pass, all integration fixtures still render unchanged.

## Notes for reviewers

- \`selfReporting\` is intentionally skipped because its collector assignment is chart-internal and not directly tied to a user-facing feature. If we wanted to lock it in too we'd want to do it on the chosen collector after \`features.selfReporting.chooseCollector\` resolves.
- The lookup checks both \`collectorValues.alloy.stabilityLevel\` and a top-level \`collectorValues.stabilityLevel\`, because \`collectorCommon.alloy.<key>\` currently lands at the top of the merged values rather than under \`.alloy\`. That's a separate chart bug (the rendered Alloy CR ends up with \`spec.stabilityLevel\` instead of \`spec.alloy.stabilityLevel\`), not addressed here — I'll send a follow-up if you want.
- The two new templates change \~110 lines; the rest of the PR diff is the test fixture.

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -5,6 +5,14 @@
 *   Add global pull policy support for the hooks and collectors (@petewall)
 *   Fix `spanMetrics.excludeDimensions` rendering: list values are now properly quoted as strings instead of unquoted bareword identifiers, which caused Alloy to fail to parse the rendered config. (#2596) (@savannahostrowski)
 *   Remove scrape protocols from the four `prometheus.operator.*` components, which do not support setting them. (@MattiasSegerdahl)
+*   Fail with a helpful message when the chart will render Alloy components that require
+    `stabilityLevel: experimental` (e.g. `global.convertClassicHistogramsToNhcb: true` or a
+    Prometheus destination on `remoteWriteProtocol: 2`) but a collector hosting those components
+    has a lower stabilityLevel, instead of crash-looping the collector at startup. Per-trigger
+    minimum stabilityLevel is declared in the `collectors.experimentalStabilityRequirements`
+    helper, so when Alloy graduates a component
+    (`experimental` → `public-preview` → `generally-available`) the validation is updated by
+    changing a single value. (@MattiasSegerdahl)
 
 ## 4.1.2
 
--- a/charts/k8s-monitoring/templates/_validations.tpl
+++ b/charts/k8s-monitoring/templates/_validations.tpl
@@ -6,6 +6,8 @@
   {{- include "destinations.validate" . -}}
   {{- include "collectors.validate.atLeastOneEnabled" . }}
   {{- include "collectors.validate.uniqueNames" . }}
+  {{- include "collectors.validate.experimentalStabilityLevel" . }}
+  {{- include "collectors.validate.experimentalStabilityLevel" . }}
 
   {{- /* Feature Config Influence */}}
   {{- $updatedValues := $.Values }}
--- a/charts/k8s-monitoring/templates/collectors/_collector_validations.tpl
+++ b/charts/k8s-monitoring/templates/collectors/_collector_validations.tpl
@@ -67,6 +67,102 @@
 {{- end }}
 {{- end }}
 
+{{/* Fails when the chart will render Alloy components that require a higher stability.level
+     than the collector hosting them is configured for. Walks enabled features, looks up each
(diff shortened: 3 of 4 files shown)


petewall:
This has been solved in this PR: https://github.com/grafana/k8s-monitoring-helm/pull/2890

PR #2605, opened 2026-05-14
fix(collectorCommon): land alloy:-subtree keys at spec.alloy on the Alloy CR

MattiasSegerdahl:
## What does this PR do?

`collectorCommon.alloy.<key>` is the chart's "settings for all Alloy collectors" knob. Today it works for some keys and silently fails for others:

| User sets | Lands at, on the Alloy CR | Operator reads it? |
|---|---|---|
| `collectorCommon.alloy.image.tag` | `spec.image.tag` | ✅ yes |
| `collectorCommon.alloy.controller.replicas` | `spec.controller.replicas` | ✅ yes |
| `collectorCommon.alloy.stabilityLevel: experimental` | `spec.stabilityLevel` | ❌ no (must be `spec.alloy.stabilityLevel`) |
| `collectorCommon.alloy.configMap.create` | `spec.configMap` | ❌ no |
| `collectorCommon.alloy.securityContext.runAsUser` | `spec.securityContext` | ❌ no |
| `collectorCommon.alloy.clustering.enabled` | `spec.clustering` | ❌ no |

The root cause is in `collector.alloy.values` (`_collector_helpers.tpl`): it does

```yaml
$userCommonValues := $.Values.collectorCommon.alloy
```

which extracts the **contents** of `collectorCommon.alloy` and then merges them at the **top of the per-collector values dict**. Whatever the user put under `collectorCommon.alloy.<key>` therefore lands at `spec.<key>` on the rendered Alloy CR. The upstream Alloy chart's values shape is hybrid: some fields live at the top (`image`, `controller`, `rbac`, `serviceAccount`, `configReloader`, `networkPolicy`, `service`, `ingress`, `extraObjects`), others under its own `alloy:` subtree (`stabilityLevel`, `configMap`, `securityContext`, `clustering`, `resources`, `logging`, `liveDebugging`, `mounts`, `lifecycle`, `livenessProbe`, `storagePath`, `listenAddr`/`listenPort`/`listenScheme`, `extraEnv`, `extraArgs`, ...). The chart's strip happens to map the first group correctly; the second group ends up at the wrong place in `spec` and the operator drops the value.

The most common way to hit this is `collectorCommon.alloy.stabilityLevel: experimental`: the chart writes it, the operator doesn't see it, and the alloy-metrics collector crash-loops on any config that requires `--stability.level=experimental` (Prometheus RWv2 protobuf, NHCB conversion, ...). The current docs and `upgrade.sh`-style snippets recommend exactly that knob.

## The fix

Split the user's `collectorCommon.alloy` dict against `collectors/upstream/alloy-values.yaml`: keys that exist under the upstream `alloy:` subtree get re-wrapped so they merge into `spec.alloy`, the rest stay at the top of `spec`.

```yaml
{{- $upstreamAlloyValues := "collectors/upstream/alloy-values.yaml" | .Files.Get | fromYaml }}
{{- $alloySubtreeKeys := keys (default dict (get $upstreamAlloyValues "alloy")) }}
{{- $rawUserCommon := default dict $.Values.collectorCommon.alloy }}
{{- $userCommonAlloy := dict }}
{{- $userCommonValues := dict }}
{{- range $key, $val := $rawUserCommon }}
  {{- if has $key $alloySubtreeKeys }}
    {{- $_ := set $userCommonAlloy $key $val }}
  {{- else }}
    {{- $_ := set $userCommonValues $key $val }}
  {{- end }}
{{- end }}
{{- if $userCommonAlloy }}
  {{- $_ := set $userCommonValues "alloy" $userCommonAlloy }}
{{- end }}
```

The list of "under-alloy" keys is sourced from the upstream Alloy chart's values, so it tracks the Alloy chart's shape as Alloy evolves. No manual list to maintain.

## Backward compatibility

The `docs/examples/private-image-registries/individual/values.yaml` and `README.md` example continue to work unchanged:

```yaml
collectorCommon:
  alloy:
    image:
      registry: my.registry.com
      repository: grafana/alloy
```

`image` is at the top of the upstream Alloy chart's values, so it stays at the top of `spec` after the split. Verified by a unit test.

`collectorCommon.alloy.controller.podAnnotations` (used in `tests/platform/aks`), `collectorCommon.alloy.liveDebugging.enabled` (`tests/platform/grafana-cloud/k8s-monitoring`), `collectorCommon.alloy.remoteConfig.*` (`tests/platform/remote-config`) — `controller`, `liveDebugging`, and `remoteConfig` are all categorised correctly by the data-driven split:

| Key | In upstream `alloy:`? | Lands at, after this PR |
|---|---|---|
| `controller` | No | `spec.controller` (unchanged) |
| `liveDebugging` | Yes | `spec.alloy.liveDebugging` (newly correct; was `spec.liveDebugging`) |
| `remoteConfig` | No (chart-internal field, not from upstream Alloy chart) | `spec.remoteConfig` (unchanged) |

The full chart suite passes (175 unit tests + 42 snapshots), including the rendered fixtures in `tests/platform/*` and `docs/examples/*`.

## Tests

`charts/k8s-monitoring/tests/collector_common_alloy_test.yaml` covers four cases:

1. `collectorCommon.alloy.stabilityLevel: experimental` → `spec.alloy.stabilityLevel: experimental` (was silently dropped)
2. `collectorCommon.alloy.image: { tag: v1.16.1, ... }` → `spec.image: { tag: v1.16.1, ... }` (preserves the existing private-image-registries example)
3. `collectorCommon.alloy.securityContext.runAsUser: 12345` → `spec.alloy.securityContext.runAsUser: 12345` (deep merge with chart defaults, was silently dropped)
4. Per-collector override: `collectors.<name>.alloy.stabilityLevel: public-preview` still wins over `collectorCommon.alloy.stabilityLevel: experimental` (precedence preserved)

## Notes for reviewers

- The diff is small (`_collector_helpers.tpl` +23 / -2, plus the test fixture and a CHANGELOG note). Most of the PR is the test + the CHANGELOG explanation.
- Naturally follows from the `stabilityLevel` validation in #2601: with that PR the chart fails loudly when the user's stabilityLevel is too low; with this PR, the `collectorCommon.alloy.stabilityLevel` escape route the user is most likely to reach for actually works.

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -1,12 +1,25 @@
 # Changelog
 
-## 4.1.3
+## Unreleased
 
-*   Add support for the write-ahead log (WAL) setting on the Loki destination (@nooboo)
-*   Add `drop` policy support for tail sampling (@petewall)
 *   Add global pull policy support for the hooks and collectors (@petewall)
 *   Fix `spanMetrics.excludeDimensions` rendering: list values are now properly quoted as strings instead of unquoted bareword identifiers, which caused Alloy to fail to parse the rendered config. (#2596) (@savannahostrowski)
 *   Remove scrape protocols from the four `prometheus.operator.*` components, which do not support setting them. (@MattiasSegerdahl)
+*   Fix `collectorCommon.alloy.<key>` propagation: keys the upstream Alloy chart hosts under its
+    own `alloy:` subtree (`stabilityLevel`, `configMap`, `securityContext`, `clustering`,
+    `resources`, etc.) now land at `spec.alloy.<key>` on the Alloy CR, where the operator reads
+    them, instead of being silently dropped at `spec.<key>`. Keys the upstream Alloy chart exposes
+    at the top of its values (`image`, `controller`, `rbac`, ...) continue to land at
+    `spec.<key>`, so existing usage like the `docs/examples/private-image-registries` example
+    (`collectorCommon.alloy.image.registry`) keeps working unchanged. (@MattiasSegerdahl)
+*   Fail with a helpful message when the chart will render Alloy components that require
+    `stabilityLevel: experimental` (e.g. `global.convertClassicHistogramsToNhcb: true` or a
+    Prometheus destination on `remoteWriteProtocol: 2`) but a collector hosting those components
+    has a lower stabilityLevel, instead of crash-looping the collector at startup. Per-trigger
+    minimum stabilityLevel is declared in the `collectors.experimentalStabilityRequirements`
+    helper, so when Alloy graduates a component
+    (`experimental` → `public-preview` → `generally-available`) the validation is updated by
+    changing a single value. (@MattiasSegerdahl)
 
 ## 4.1.2
 
--- a/charts/k8s-monitoring/templates/collectors/_collector_helpers.tpl
+++ b/charts/k8s-monitoring/templates/collectors/_collector_helpers.tpl
@@ -116,8 +116,29 @@ app.kubernetes.io/instance: {{ include "collector.alloy.fullname" . }}
 {{- define "collector.alloy.values" }}
 {{- /* The default settings set for all Alloy instances by this chart */}}
 {{- $defaultValues := "collectors/alloy-values.yaml" | .Files.Get | fromYaml }}
-{{- /* Settings in values.yaml for all Alloy instances */}}
-{{- $userCommonValues := $.Values.collectorCommon.alloy }}
+{{- /* Settings in values.yaml for all Alloy instances. `collectorCommon.alloy.<key>` mixes two
+       shapes of keys: ones that the upstream Alloy chart exposes at the top of its values
+       (image, controller, rbac, serviceAccount, ...) and ones that it nests under its own
+       `alloy:` subtree (stabilityLevel, configMap, securityContext, clustering, resources, ...).
+       Splitting the user-provided dict against `collectors/upstream/alloy-values.yaml` lets the
+       first group keep landing at `spec.<key>` on the Alloy CR (preserves the existing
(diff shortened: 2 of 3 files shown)


petewall:
I think I see the issue.
```yaml
collectorCommon:
  alloy: {...}
```
Is meant to match the the top level of the alloy instance. That means that we will often do stuff like:

```yaml
collectorCommon:
  alloy:
    alloy:
      stabilityLevel: ...
```

The reason for the "alloy" inside of "collectorCommon" is that it allows for a future where we could have:

```yaml
collectorCommon:
  otelCollector: {...} # Settings that modify all otel collectors
  alloy: {... } # Settings that modify all alloy collectors
```

MattiasSegerdahl:
Ah, that makes complete sense — and the `otelCollector` future-proofing is a strong reason to keep it as-is. I had the wrong mental model: I read `collectorCommon.alloy` as "settings that get merged into the `spec.alloy` subtree" rather than "settings for the top of an alloy collector instance (where `alloy:` itself is one subtree among many)". With your framing the existing behavior is internally consistent and this PR is solving a non-bug. Closing it.

(Quick sanity check on `main` against your framing — works as advertised:
```yaml
collectorCommon:
  alloy:
    alloy:
      stabilityLevel: experimental
```
renders `spec.alloy.stabilityLevel: experimental` on every Alloy CR. So the example I should have led with in `docs/examples` is `collectorCommon.alloy.alloy.<key>` for the under-`alloy:` cases, not the change in this PR.)

One small DX thing I keep tripping over and that the broader user-visible issue I tried to fix here actually stems from: it's easy to type `collectorCommon.alloy.stabilityLevel` (one `alloy:` short of correct) and get a silently-no-op result. Would a chart-level validation that catches "you put a known-`spec.alloy.*` key at `collectorCommon.alloy.<key>` — did you mean `collectorCommon.alloy.alloy.<key>`?" be welcome as a small follow-up? Same idea as #2601 (also still open), just for this specific footgun. Happy to skip if you'd rather just rely on documentation.

PR #2615, opened 2026-05-18
fix: treat empty service attributes as missing

cedricziel:
## Summary
- treat empty `service.name` and `service.namespace` values as missing in the pod logs via OpenTelemetry transform so Kubernetes-derived fallbacks still populate them
- update OTLP attribute-to-resource promotion to overwrite empty destination resource attributes instead of only `nil` ones
- require non-empty source values before copying during fallback and promotion

## Validation
- rendered the affected chart fixtures with `make tests/platform/otlp-gateway/.rendered/output.yaml tests/platform/grafana-cloud/k8s-monitoring/.rendered/output.yaml tests/platform/grafana-cloud/app-observability/.rendered/output.yaml`


rlankfo (dismissed):
lgtm

petewall (approved):

PR #2618, opened 2026-05-21
feat(k8s-monitoring-v1): add configurable Prometheus OTLP timeout

wjma90:
feat(k8s-monitoring-v1): add configurable Prometheus OTLP timeout

Add externalServices.prometheus.otlpTimeout to configure the otelcol.exporter.otlp metrics_service timeout when Prometheus uses the otlp protocol.
When unset, the chart keeps the current behavior and lets Alloy use its default timeout.


petewall:
I brought this in with #2621
I'll make a new release soon.

PR #2643, opened 2026-05-28
Update Alloy Operator, OpenCost, and Windows Exporter

petewall:


petewall:
Removed kube-state-metrics from the upgrade list, because it broke its own sharding.

petewall:
Making this a draft until we get a new version of alloy-operator out, too

rlankfo (approved):

Issue #2547, opened 2026-04-30
Align `service.name` detection across logs, profiles, and Beyla with OTel Operator semconv


petewall:
# Align `service.name` / `service.namespace` detection across signals with Beyla / OTel Operator semconv

## Summary

The chart derives `service.name` and `service.namespace` for logs and profiles from a ranked list of pod metadata, but the lists are **inconsistent across signals** and **none of them match what Beyla produces** for the same pod. As a result, the same workload shows up under different `service.name` values in metrics vs. logs vs. profiles, which breaks correlation in Grafana (Application Observability, IHub, etc.).

This was first noticed while debugging why correlation didn't work out-of-the-box for the Istio bookinfo demo: Beyla emits `target_info{service_name="reviews-v2"}` (the Deployment owner name), while logs and profiles for the same pod report `service.name="reviews"` (the container name).

## Current state

| Signal | File | Ranked list |
|---|---|---|
| Logs (Loki) | `charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl` (L89-102) | `resource.opentelemetry.io/service.name` annotation → `app.kubernetes.io/name` label → **container name** |
| Logs (OTel) | `charts/k8s-monitoring/charts/feature-pod-logs-via-opentelemetry/templates/_module.alloy.tpl` (L120-128) | `app.kubernetes.io/name` → `k8s.deployment.name` → `k8s.replicaset.name` → `k8s.statefulset.name` → `k8s.daemonset.name` → `k8s.cronjob.name` → `k8s.job.name` → `k8s.pod.name` → `k8s.container.name` *(no annotation, no `instance` label)* |
| Profiles (eBPF) | `charts/k8s-monitoring/charts/feature-profiling/templates/_ebpf.tpl` (L84-101) | annotation → `app.kubernetes.io/instance` → `app.kubernetes.io/name` → **container name** |
| Profiles (Java) | `charts/k8s-monitoring/charts/feature-profiling/templates/_java.tpl` (L111-127) | same as eBPF |
| Profiles (pprof) | `charts/k8s-monitoring/charts/feature-profiling/templates/_pprof.tpl` (L72-87) | same as eBPF |
| Metrics (Beyla `target_info`) | Beyla internal | annotation → `app.kubernetes.io/instance` → `app.kubernetes.io/name` → **owner workload name** → pod → container *(per OTel Operator k8s semconv)* |

### Inconsistencies

[...]

Issue #1540, opened 2025-05-13
[pod logs] Use the K8s `deployment/cronjob/job` name in the `service_name` logic


cyrille-leclerc:
We want to align the `service_name` logic of the [K8s Monitoring Helm Chart Pod Logs feature](https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring/charts/feature-pod-logs ) on the logic used by the OpenTelemetry Operator.

For this, we need to retrieve the `deployment/cronjob/job` of the pods through the "controller-by" attribute.
If the OpenTelemetry Collector K8s Processor is capable of retrieving these metadata, it's not clear if Alloy's `discovery.relabel` can also do it.

* https://github.com/grafana/k8s-monitoring-helm/blob/v2.0.28/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_pod_discovery.alloy.tpl
* The source code of the logic to etrive pod metadata seems to be https://github.com/prometheus/prometheus/blob/main/discovery/kubernetes/pod.go

## OpenTelemetry Operator logic for `service.name`

First value found:
- `pod.annotation[resource.opentelemetry.io/service.name]`
- `if (config[useLabelsForResourceAttributes]) pod.label[app.kubernetes.io/name]`
- `k8s.deployment.name`
- `k8s.replicaset.name`
- `k8s.statefulset.name`
- `k8s.daemonset.name`
- `k8s.cronjob.name`
- `k8s.job.name`
- `k8s.pod.name`
- `k8s.container.name`

cyrille-leclerc:
There may be a hack looking at `__meta_kubernetes_pod_controller_kind` (e.g. `ReplicaSet`) and `__meta_kubernetes_pod_controller_name` (e.g. `orders-prometheus-postgres-exporter-6f65896bd5`) and extracting the `substringBeforeLast(__meta_kubernetes_pod_controller_name, '-')`.


```
{
	__address__                                              = "10.0.12.114:9187",
	__meta_kubernetes_namespace                              = "ditl-demo-prod",
	__meta_kubernetes_pod_annotation_checksum_config         = "01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b",
	__meta_kubernetes_pod_annotation_checksum_secrets        = "4d8241978bef61f991832cebf24c44d8aad147d9a9629663a01a24c05ec414d7",
[...]

Issue #942, opened 2024-11-22
Mismatch between Beyla and Loki pod logs on `service.name`, `service.namespace`, and `service.instance.id`


cyrille-leclerc:
# Problem description

There is inconsistency in critical resource attributes of the telemetry of applications instrumented with Beyla for metrics & traces combined with the Grafana K8s Helm Chart for logs emitted through k8s stdout.
This inconsistency, different `service.name`, `service.namespace`, and `service.instance.id` break correlations capabilities in Grafana Cloud, particularly in Grafana Application Observability.

# Example of inconsistencies

  | TRACE BY BEYLA | LOGS BY GRAFANA K8S MONITORING HELM CHART
-- | -- | --
host.id | 8b3ab3487d2d4b3d9ffa0b93accc6b78 |  
host.name | authentication-deployment-7d77f7886b-6kwhn |  
job |   | dev/authentication-container
k8s.cluster.name | dev |  dev
k8s.container.name |  | authentication-container
k8s.deployment.name | authentication-deployment |  
k8s.namespace.name | dev |  dev
k8s.node.name | atlubun2devn2 |  
k8s.pod.name | authentication-deployment-7d77f7886b-6kwhn |  
k8s.pod.start_time | 2024-11-07 19:53:02 +0000 UTC |  
k8s.pod.uid | 4caaca67-4386-4148-919c-1615e823cece |  
k8s.replicaset.name | authentication-deployment-7d77f7886b |  
otel.library.name | github.com/grafana/beyla |  
pod |   | authentication-deployment-7d77f7886b-6kwhn
🔴  service.instance.id | beyla-ld92z-2310310 |  
🔴  service.name | authentication-deployment |  authentication-container
🔴  service.namespace | dev |  
telemetry.sdk.language | go |  
telemetry.sdk.name | beyla |  

----

<img height="150px" src="https://github.com/user-attachments/assets/acf9ef4f-ca95-43f8-b21c-f49f581c6a54" />

---

<img height="150px" src="https://github.com/user-attachments/assets/5577fffd-4a5e-4758-a2d2-0a96e0e98366" />

# Root Cause Analysis - Inconsistent naming strategies

<table>
<tr>   
   <th>Resource Attribute</th>
   <th>OTel Operator</th>
   <th>Beyla K8s</th>
   <th>Alloy K8s + Loki</th>
</tr>

<tr>   
   <th><code>service.name</code></th>
   <td><code>first_non_null(
[...]

PR #2651, opened 2026-06-01
feat: opt-in Kubernetes service attribute fallback for App Observability signals

jmichalek132:
## Summary

Reworked per review. Adds an opt-in flag that fills missing or empty `service.name` / `service.namespace` resource attributes from Kubernetes metadata on metrics, logs, and traces in the Application Observability pipeline. Part of #2547, related #942, #1540.

## Changes

- New `applicationObservability.processors.transform.setServiceAttributesFromKubernetes` (default `false`). When enabled, k8sattributes extracts the `app.kubernetes.io/name` pod label into an internal attribute, the transform fills `service.name` (label -> deployment -> statefulset -> daemonset -> cronjob -> job -> pod) and `service.namespace` (namespace) on all three signals, then deletes the internal attribute after user-provided resource transforms. With the flag off, the Application Observability render is unchanged from main.
- No replicaset/container rules: the k8sattributes metadata defaults don't extract them (dead code).
- Fix Pod Logs via OpenTelemetry: the chain read `attributes["app.kubernetes.io/name"]` but the label is extracted as `app_kubernetes_io_name`, so the rule never fired. Now derived from `labels`, omitted when unmapped. Also trims the unreachable replicaset/container rules. Unconditional bug fix.
- CHANGELOG entries plus a precedence note in the OTLP destination docs (receiver-side fallback wins over `logToResource`/`datapointToResource` promotion).
- ArgoCD test Makefile change retained so the ArgoCD deployment-alternative integration test passes on fork PRs.

## Validation

- New helm-unittest coverage in the feature subcharts and parent chart: flag off/on, ordering (setters -> user transforms -> cleanup), custom labels, pod-logs mapping variants.
- `make -C charts/k8s-monitoring unittest examples integration-test-checks platform-test-checks HELM_VERSION=4.1.4`: 39 suites, 180 tests, green and idempotent; subchart suites run separately, also green.
- Flag-enabled rendered Alloy config passes `alloy fmt` and `alloy validate`.


jmichalek132:
On the removal of `{{- if .Values.traces.transforms.resource }}`:

This is intentional. Previously that guard only rendered trace resource statements when users configured custom trace resource transforms. The new service fallback statements are built-in defaults, so keeping the guard would make the fix inactive for the default Application Observability trace path.

The block remains gated by `{{- if .Values.traces.enabled }}`, so it only renders when traces are enabled. User-provided `.Values.traces.transforms.resource` statements are still appended after the built-in fallback statements.

petewall (commented):
This is great and I want this change. A few things that we'll need to address, though.

We don't currently capture the `app.kubernetes.io/name` label. Ideally, we'd add a:
```yaml
processors:
  k8sattributes:
    labels:
      - app.kubernetes.io/name
```
Which would behave like the metadata section does now, but extract that label as an attribute.

Next, we could probably leave `k8s.replicaset.name` and `k8s.container.name` off the list, since we don't capture them and they're likely covered by other attributes earlier.

Would we want a flag in values.yaml to enable or disable this functionality? It's changing behavior that some might not expect. I might target this for the upcoming 4.2.

Finally, can we add a line in CHANGELOG.md for this?

jmichalek132:
Thanks for the review @petewall - reworked along your points:

1. `app.kubernetes.io/name` is now extracted via k8sattributes when the feature is on, into an internal attribute that's deleted after user-provided resource transforms, so the exported schema doesn't change and user `labels` overrides can't drop it.
2. replicaset/container dropped. (The OTel Operator semconv includes them, but our metadata defaults don't extract those fields, so they'd be dead rules.)
3. Flag added, default `false`, and it now covers metrics and logs too, so all three signals get consistent service identity (part of #2547).
4. CHANGELOG added.

Bonus fix: the Pod Logs via OpenTelemetry chain read `attributes["app.kubernetes.io/name"]` but the label is extracted as `app_kubernetes_io_name`, so the app-label rule never fired. Fixed unconditionally with its own CHANGELOG entry.

jmichalek132:
hey @petewall any updates on reviewing this pr :) ?

PR #2665, opened 2026-06-03
Add extraDiscoveryRules to Service Integrations

bradleypettit:
## What this does

Adds an `extraDiscoveryRules` setting to the Service Integrations feature, letting users inject custom **pre-scrape** `discovery.relabel` rules into an integration's metrics discovery. The rules run inside the same `discovery.relabel` component as the built-in relabeling, so all `__meta_*` labels are still present — enabling, for example, a `labelmap` from pod annotations to enrich integration metrics with cost-attribution labels.

## Why

Integrations previously exposed only `labelSelectors` / `fieldSelectors` / `namespaces` / `metrics.tuning` (plus post-scrape `extraMetricProcessingRules` on a few types). None of those can map discovery-time `__meta_*` labels (e.g. `__meta_kubernetes_pod_annotation_*`), because `extraMetricProcessingRules` runs post-scrape where those labels no longer exist. This closes that gap and mirrors the existing `extraDiscoveryRules` already offered by `annotationAutodiscovery`, `clusterMetrics`, `costMetrics`, and `hostMetrics`.

## Scope

Added to all nine Kubernetes service-discovery integrations: **alloy, cert-manager, dcgm-exporter, etcd, grafana, istio, loki, mimir, tempo**.

- Most are **type-level** (`integrations.<name>.extraDiscoveryRules`), applying to every instance of that integration.
- **istio** exposes it **per-section** (`sidecarMetrics.extraDiscoveryRules` and `istiodMetrics.extraDiscoveryRules`) to match how istio already splits `labelSelectors`, `tuning`, and `extraMetricProcessingRules`, and because the two scrape different roles (`pod` vs `endpointslice`) with different `__meta_*` labels.
- **mysql / postgresql are intentionally excluded** — they scrape `prometheus.exporter.*` targets with no `__meta_*` discovery labels; they already expose `extraMetricProcessingRules` for post-scrape enrichment.

## Example

```yaml
integrations:
  alloy:
    extraDiscoveryRules: |-
      rule {
        action = "labelmap"
        regex  = "__meta_kubernetes_pod_annotation_resource_opentelemetry_io_(.*)"
      }
    instances:
      - name: alloy
        labelSelectors:
          app.kubernetes.io/name: [alloy-metrics]
```

## Testing

- `make build` regenerated the schema and docs.
- `helm lint` passes; `helm unittest` passes (136 tests / 43 snapshots), with added coverage for the declare-module pattern (alloy), the per-instance pattern (cert-manager), and the istio sidecar/istiod split (new suite).
- Default (unset) output is byte-identical to before — no snapshot churn.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


petewall (dismissed):
This is great, but I'd love to have the `extraDiscoveryRules` defined in the values files for the other integrations, too. It's added in the istio ones, but not the other integrations.

They should be within the "instance" of the integration, because those are discovering unique targets.

bradleypettit:
> This is great, but I'd love to have the `extraDiscoveryRules` defined in the values files for the other integrations, too. It's added in the istio ones, but not the other integrations.
> 
> They should be within the "instance" of the integration, because those are discovering unique targets.

It's funny I actually designed it that way initially, and then reworked it to be once per integration and not once per integration instance. Mainly because the templating in this case was a little cleaner and easier to read. Istio was a bit of a special case in comparison (it already is).

Both approaches work fine for me so can refactor it to be per instance.

bradleypettit:
@petewall updated - does this suit better for what you're suggesting?

TylerHelmuth (dismissed):

TylerHelmuth (approved):

PR #2700, opened 2026-06-12
Detect and report platform

petewall:
Previously, we had some platform detection, but it was for checking for invalid configurations. For example, if your cluster is OpenShift, but we're not declaring the `openshift` platform, we error and tell them to fix it.

Now, we're going to try and detect any platform that we're deploying to, then use that in the self-reporting metric and as an attribute to remote config settings.

### DIFF
--- a/charts/k8s-monitoring/templates/_platform_validations.tpl
+++ b/charts/k8s-monitoring/templates/_platform_validations.tpl
@@ -1,19 +1,69 @@
-{{ define "validations.platform" -}}
+{{- /*
+Detects the platform this cluster is running on by inspecting the available API
+versions and the metadata of the cluster's nodes. Returns the detected platform
+name ("openshift", "aks", "gke", or "eks"), or an empty string if the platform
+could not be determined. Input: . (root context)
+*/ -}}
+{{ define "validations.platform.detect" -}}
   {{- $platform := "" }}
-  {{- if (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") }}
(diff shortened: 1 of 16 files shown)


charts/k8s-monitoring/templates/_platform_validations.tpl
```diff
@@ -1,20 +1,72 @@
...
+name ("openshift", "aks", "gke", or "eks"), or an empty string if the platform
+could not be determined. Input: . (root context)
+*/ -}}
+{{ define "validations.platform.detect" -}}
   {{- $platform := "" }}
-  {{- if (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") }}
-    {{- $platform = "openshift" }}
-    {{- include "validations.platform.openshift" . }}
+  {{- if .Capabilities }}
+    {{- if (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") }}
```
TylerHelmuth: I'm not sure what our typical style is for this chart, but we could drop the first if statement and do 

```
{{- if ((.Capabilities).APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") }}
```

TylerHelmuth (dismissed):

charts/k8s-monitoring/templates/_platform_validations.tpl
petewall: Whoa, I didn't know you could do that!

TylerHelmuth (approved):

PR #2720, opened 2026-06-18
The timeout placement is different between otlp and otlp http

petewall:
Unfortunately, due to inconsistencies, we need another fix for the OTLP destination timeout placement...

In `otelcol.exporter.otlp` ([docs](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.otlp/)):
```
otelcol.exporter.otlp "<LABEL>" {
  timeout = "60s"
  client {
    ...
  }
}
```

In `otelcol.exporter.otlphttp` ([docs](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.otlphttp/)):
```
otelcol.exporter.otlphttp "<LABEL>" {
  client {
    timeout = "60s"
    ...
  }
}
```

### DIFF
--- a/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
+++ b/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
@@ -459,6 +459,9 @@ otelcol.processor.memory_limiter {{ include "helper.alloy_name" $.destinationNam
 
 {{- if eq .protocol "grpc" }}
 otelcol.exporter.otlp {{ include "helper.alloy_name" $.destinationName | quote }} {
+{{- if .timeout }}
+  timeout = {{ .timeout | quote }}
+{{- end }}
 {{- else if eq .protocol "http" }}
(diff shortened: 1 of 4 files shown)


charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
TylerHelmuth, line 1: I'm not good at reading the grafana configuration syntax yet, but if the underlying components are the opentelemetry otlpexporter and otlphttpexporter, then they both have `timeout` in the same spot:

```yaml
exporters:
  otlp:
    endpoint: <an endpoint>
    timeout: 10s

  otlphttp:
    endpoint: <an endpoint>
    timeout: 10s
```

TylerHelmuth: I reviewed the component structs and confirmed the collector yaml works like this because of the unmarshalling, but the actual timeout fields in the config struct do live in different spots.

TylerHelmuth (approved):

PR #2735, opened 2026-06-24
Fix Beyla's SCC on OpenShift to include hostPorts

### LINKED ISSUE #2734: autoInstrumentation on OpenShift deploys scc/k8s-monitoring-beyla without allowHostPorts
I enable beyla on OpenShift 4.20 in the chart version 4.1.6 with:

```
  autoInstrumentation:
    enabled: true
    collector: alloy-metrics
    destinations:
      - central-otlp
    beyla:
      contextPropagation:
        enabled: true
      deliverTracesToApplicationObservability: true
      preset: application
      ...
```

The daemonset fails to create pods with:

```
Warning   FailedCreate        daemonset/dev2-k8s-monitoring-beyla             Error creating: pods "dev2-k8s-monitoring-beyla-" is forbidden: unable to validate against any security context constraint: [...provider dev2-k8s-monitoring-beyla: .containers[0].containers[0].hostPort: Invalid value: 9090: Host ports are not allowed to be used]
```

The scc should set `allowHostPorts: true`

petewall:
Fixes #2734

### DIFF
--- a/charts/k8s-monitoring/charts/feature-auto-instrumentation/templates/platform_specific/openshift/beyla-scc.yaml
+++ b/charts/k8s-monitoring/charts/feature-auto-instrumentation/templates/platform_specific/openshift/beyla-scc.yaml
@@ -8,7 +8,7 @@ allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowHostNetwork: true
 allowHostPID: true
-allowHostPorts: false
+allowHostPorts: true
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
 allowedCapabilities: []
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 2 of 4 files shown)


charts/k8s-monitoring/charts/feature-auto-instrumentation/templates/platform_specific/openshift/beyla-scc.yaml
```diff
@@ -8,7 +8,7 @@ allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowHostNetwork: true
 allowHostPID: true
-allowHostPorts: false
+allowHostPorts: true
```
TylerHelmuth, line 11: Does beyla always need host ports or only when some specific capabilities are enabled?

petewall: Good question... let me take a look

petewall: Without going into too much detail, OpenShift considers this a hostPort whenever the Beyla daemonset uses hostNetwork *and* a container port (which effectively, the combination makes it a hostPort).

We could dive in and make it conditional, but we're not doing that for the other things like hostNetwork or hostPID, so I think we just match and call it a day for now.

TylerHelmuth (approved):

PR #2738, opened 2026-06-24
feat: start verifying charts before use

TylerHelmuth:
Update k8s-manifest-tail to the latest version.

The latest version is signed, so also add support for verifying the dependency before using it.


keys/grafana-helm-charts-pubkey.gpg
petewall, line 1: Can we include a Makefile target, either here or in the top-level Makefile, to pull the public key from 1password and save here?

petewall (commented):
So this only works for specific charts we know are signed?
Any thing we can detect if it's verifiable or not?

keys/grafana-helm-charts-pubkey.gpg
TylerHelmuth: Good idea

Makefile
```diff
@@ -33,6 +33,11 @@ clean: ## Clean all charts
 build: check-helm-version ## Build all charts
 	make -C charts/k8s-monitoring $@;
 
+##@ Keys
+.PHONY: update-signing-keys
+update-signing-keys: ## Refresh signing keys in keys/ from 1Password (requires the op CLI)
+	op --account grafana.1password.com read "op://Helm Maintainers/Helm Chart Signing Key/gpg-public-key.asc" | gpg --dearmor > keys/grafana-helm-charts-pubkey.gpg
```
petewall, line 41:
```suggestion
.PHONY: update-signing-keys
update-signing-keys: keys/grafana-helm-charts-pubkey.gpg ## Refresh signing keys in keys/ from 1Password (requires the op CLI)

keys/grafana-helm-charts-pubkey.gpg:
	op --account grafana.1password.com read "op://Helm Maintainers/Helm Chart Signing Key/gpg-public-key.asc" | gpg --dearmor > keys/grafana-helm-charts-pubkey.gpg
```

This should prevent it from re-fetching over and over.

petewall (dismissed):

petewall (dismissed):

petewall:
the unit test "check generated files" is failing because the updated k8s-manifest-tail will also update some of the generated examples. Run `make clean build test` to get things rebuilt.

Also, put a note in the Unreleased section of CHANGELOG that says that we're updating k8s-manifest-tail to 0.1.5

petewall (approved):
🚀

PR #2740, opened 2026-06-25
feat: verify prometheus-community charts

TylerHelmuth:
Add support for verifying the prometheus community charts


petewall (approved):
When we eventually have our public key published to the Helm charts, Rebo, we should update the file to pull from there rather than 1Password

Issue #1268, opened 2025-02-21
v2 Kubelet feature - wrong value for discovery.relabel


cristian-1:
Hi,
there is a typo for the extraDiscoveryRules block in the kubelet feature.
Both values file and documentation say to use "extraDiscoveryRules" to add additional rule blocks in discovery.relabel while the template looks for .Values.kubelet.extraRelabelingRules.

Docs: https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring/charts/feature-cluster-metrics#kubelet

values: https://github.com/grafana/k8s-monitoring-helm/blob/e60649459d0427b1aaa6a29657f6f6d97ebe61e1/charts/k8s-monitoring/charts/feature-cluster-metrics/values.yaml#L63

template: https://github.com/grafana/k8s-monitoring-helm/blob/e60649459d0427b1aaa6a29657f6f6d97ebe61e1/charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet.alloy.tpl#L58

petewall:
Yes! Thank you for spotting this!

petewall:
@rlankfo , can you fix this and release 2.0.13?

rlankfo:
fixed in 2.0.13 👍

PR #2744, opened 2026-06-26
fix: honor extraDiscoveryRules for kubeletResource and kubeletProbes

### LINKED ISSUE #2743: [BUG] kubeletResource.extraDiscoveryRules and kubeletProbes.extraDiscoveryRules are silently ignored
### What's wrong

`kubeletResource.extraDiscoveryRules` and `kubeletProbes.extraDiscoveryRules`
are documented in the README and defined in `values.yaml`, but are silently
ignored — any rules supplied never appear in the rendered Alloy config.

The templates reference a values key that does not exist
(`extraRelabelingRules`) instead of the documented one
(`extraDiscoveryRules`), so the `{{- if }}` guard is always false and the
user-supplied rules are dropped.

This is the same typo as #1268, which was fixed for the `kubelet` template but
never propagated to these two sibling templates.

### Affected files

- `charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet_resource.alloy.tpl` (lines 37-39)
- `charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet_probes.alloy.tpl` (lines 37-39)

Both reference `.Values.kubeletResource.extraRelabelingRules` /
`.Values.kubeletProbes.extraRelabelingRules` inside their `discovery.relabel`
block. `extraRelabelingRules` is not defined anywhere in `values.yaml` or
`values.schema.json`. The correct key — `extraDiscoveryRules` — is what every
other cluster-metrics component (`kubelet`, `cadvisor`, `apiServer`,
`kubeProxy`, `kubeScheduler`, `kubeControllerManager`, `kubeDNS`,
`kube-state-metrics`) uses in its `discovery.relabel` block.

### To reproduce

```yaml
clusterMetrics:
  kubeletResource:
    extraDiscoveryRules: |
      rule {
        source_labels = ["__meta_kubernetes_node_label_topology_kubernetes_io_region"]
        target_label  = "region"
      }
```

Render the chart — the rule does not appear in the
`discovery.relabel "kubelet_resources"` block.

### Expected

Rules supplied via `extraDiscoveryRules` are added to the `discovery.relabel`
component, matching the README and the behaviour of all sibling components.

### Fix

Rename the `extraRelabelingRules` reference to `extraDiscoveryRules` in both
templates. (The `extraMetricProcessingRules` reference in the
`prometheus.relabel` block is already correct.)

### Version

main (also affects current release)

bradleypettit:
## What this does

Fixes a typo that caused `kubeletResource.extraDiscoveryRules` and
`kubeletProbes.extraDiscoveryRules` to be silently ignored.

Both templates referenced a values key that does not exist
(`extraRelabelingRules`) inside their `discovery.relabel` blocks, so the
`{{- if }}` guard was always false and user-supplied rules were dropped. This
renames the reference to the documented key `extraDiscoveryRules`, matching
every sibling cluster-metrics component (`kubelet`, `cadvisor`, `apiServer`,
`kubeProxy`, `kubeScheduler`, `kubeControllerManager`, `kubeDNS`,
`kube-state-metrics`).

This is the same typo as #1268, which was fixed for the `kubelet` template but
never propagated to these two siblings.

## Affected files

- `charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet_resource.alloy.tpl`
- `charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet_probes.alloy.tpl`

The `extraMetricProcessingRules` reference in the `prometheus.relabel` block was
already correct and is unchanged. The README already documents the correct key,
so no docs regeneration is needed.

Fixes #2743

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+*   Fix `kubeletResource.extraDiscoveryRules` and `kubeletProbes.extraDiscoveryRules` being silently dropped. (#2743) (@bradleypettit)
 *   Surface `deploy: false` defaults for `telemetryServices.beyla` and `telemetryServices.sdkInjector` in the top-level values, matching the other telemetry services. This prevents them from deploying unintentionally on Helm versions affected by a subchart value coalescing regression (helm/helm#32132). (#2781, #2782) (@petewall)
 *   Change the default remote configuration `pollFrequency` from `5m` to `30s`. (@TylerHelmuth)
 *   Add `CLUSTER_NAME` environment variable to remote-config-enabled collectors alongside `NAMESPACE` and `POD_NAME`. (#2775) (@petewall)
--- a/charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet_probes.alloy.tpl
+++ b/charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_kubelet_probes.alloy.tpl
@@ -34,8 +34,8 @@ discovery.relabel "kubelet_probes" {
     target_label  = "__metrics_path__"
   }
 {{- end }}
-{{- if .Values.kubeletProbes.extraRelabelingRules }}
-{{ .Values.kubeletProbes.extraRelabelingRules | indent 2 }}
+{{- if .Values.kubeletProbes.extraDiscoveryRules }}
+{{ .Values.kubeletProbes.extraDiscoveryRules | indent 2 }}
 {{- end }}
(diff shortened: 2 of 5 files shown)


TylerHelmuth (commented):
@bradleypettit can you update https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/charts/feature-cluster-metrics/tests/custom_rules_test.yaml with this scenario? Also please add a changelog entry for this bug fix

petewall (approved):

TylerHelmuth (approved):

jewbetcha (approved):

PR #2760, opened 2026-06-30
K8s injection controller

grcevski:
# Description

Adds support for deploying the k8s sdk injection controller with appropriate RBAC permissions for Alloy to be able to write the config maps.

I made the change with couple of assumptions:
- Since we need to mark which service account is writing the config map for the injector, I pinned it to `grafana-k8s-monitoring-alloy-metrics`. I couldn't figure out how to retrieve the actual name.
- The Alloy PR where enabling the config map RBAC is still in review, so I decided to add custom RBAC for this purpose. I couldn't reuse the rbac.rules from the Alloy helm chart, because they don't seem to support setting namespace for the individual rule. Doing it that way would open up the new RBAC to the full cluster, which isn't desirable for the newly added RBAC.

## Authorship


charts/k8s-monitoring/values.yaml
```diff
@@ -501,6 +501,34 @@ telemetryServices:
...
+
+    # -- Render a Namespace object for `namespace.name`.
+    # @section -- SDK Injector
+    create: false
+
+  # -- Comma-separated allowlist of usernames permitted to create or update annotated injection ConfigMaps. By default,
+  # Kubernetes expands `$(POD_NAMESPACE)` to the namespace where the SDK Injector pod is running.
+  # We add grafana-k8s-monitoring-alloy-metrics as the service account allowed to write the config maps.
+  # @section -- SDK Injector
+  allowedConfigMapWriters: "system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-metrics"
```
mariomac, line 530: If I understand correctly, this will only work if the deployment name is explicitly `grafana-k8s-monitoring` and will fail with some other examples I've seen in the code, like `k8smon-alloy-metrics`.

Will see if I can find a way to dynamically get this name from somewhere else.

mariomac: For a quick merge, I guess we can always document this behavior and let the users just to override this value.

charts/k8s-monitoring/templates/_sdk_injector_extra_objects.tpl
```diff
@@ -0,0 +1,53 @@
...
+Do not use Alloy `spec.rbac.rules` for this temporary permission. Without
+`spec.rbac.namespaces`, Alloy renders those rules as cluster-scoped RBAC; setting
+`spec.rbac.namespaces` would scope every Alloy RBAC rule, not just this SDK
+Injector ConfigMap writer rule. Rendering a separate Role/RoleBinding keeps only
+this permission scoped to the namespace where the Alloy collector is deployed.
+*/}}
+{{- define "sdkInjector.extraObjects" -}}
+{{- $sdkInjector := index .Values "sdkInjector" | default dict }}
+{{- if $sdkInjector.enabled }}
+{{- range $collectorName := include "collectors.list.enabled" . | fromYamlArray }}
```
mariomac, line 21: Claude says:

>  RBAC granted to every collector, but only one is intended (altitude / least-privilege)
> The range over collectors.list.enabled renders a configmap get/create/update Role+Binding for every enabled Alloy collector (alloy-logs, alloy-singleton, …), yet only the metrics collector is in the allowlist and actually needs it. This over-grants configmap write to collectors that never use it. Consider scoping the shim to the specific collector(s) that write injection ConfigMaps rather than all of them.

I'm sure there is a context explaining the current implementation so ATM I'll just leave as this and let's see if we can refine lter.

mariomac (commented):
Taking over the task. Leaving some previous comments as "pending aspects" that we can treat in future PRs.

mariomac:
I don't have permissions to push to this PR so I'll continue it in https://github.com/grafana/k8s-monitoring-helm/pull/2767

petewall:
Rolled these changes into https://github.com/grafana/k8s-monitoring-helm/pull/2765

PR #2765, opened 2026-06-30
Instrumentation hub compatibility

petewall:
# Description

Enables full compatibility with Instrumentation Hub:

* Adds Beyla K8s Cache inside of `telemetryServices`
  * This *cannot* be used to deploy Beyla itself (that would be a breaking change)
  * We have validators to protect things
* Adds the SDK Injection controller to `telemetryServices`
  * Validators will check and tell users how to set the service account reference
* Adds a `service-discovery` collector preset, which sets the `service-discovery=true` FM attribute

An instrumentation-hub platform test should exercise all of these and will eventually run queries to Grafana Cloud to ensure it's working end-to-end.

Fixes #

## Authorship


petewall:
One of these changes is to put the Beyla helm chart inside of `telemetryServices`. It wouldn't remove the one inside of `autoInstrumentation`, because that would be a breaking change.

mariomac (approved):
Good to me! Thanks!

charts/k8s-monitoring/charts/telemetry-services/templates/_validate.tpl
```diff
@@ -1,6 +1,85 @@
...
+
+For now, the Beyla subchart under telemetryServices is only meant to deploy the standalone Kubernetes metadata
+cache (k8sCache), not Beyla itself. Beyla should be deployed by the Auto-Instrumentation feature.
+This check:
+  1. Blocks enabling the Beyla agent here (telemetryServices.beyla.enabled), pointing users at the
+     autoInstrumentation feature instead.
+  2. Prevents deploying two k8sCache instances when both this service and an enabled autoInstrumentation
+     feature would each create one.
+*/}}
+{{- define "telemetryServices.validate.beyla" }}
```
TylerHelmuth, line 55: We have a bug where if the telemetryServices beyla and autoinstrumenation beyla are both configured they will conflict on the serviceaccount, clusterrole, and clusterrolbinding name. Example values.yaml that produces the issue:

```yaml
cluster:
  name: collision-test

destinations:
  localPrometheus:
    type: prometheus
    url: http://prometheus-server.prometheus.svc:9090/api/v1/write

autoInstrumentation:
  enabled: true
  collector: alloy-metrics
  beyla:
    k8sCache:
      replicas: 0

telemetryServices:
  beyla:
    deploy: true
    k8sCache:
      replicas: 1

collectors:
  alloy-metrics:
    enabled: true
```

TylerHelmuth: I'll add a `nameOverride` like the sdkinjector values.yaml

TylerHelmuth (approved):

mariomac (changes requested):
redacted to reduce noise

mariomac (approved):
Ok. I managed to get the injector working in this PR. What I had to do:

Set:
```
telemetryServices:
  sdkInjector:
    deploy: true
    allowedConfigMapWriters: system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-logs,system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-metrics,system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-profiles,system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-receiver,system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-singleton
  beyla:
    deploy: true
```

But I would had expected that any of the alloy instances already included a beyla.ebpf receiver defining at least something like:

```hcl
      injector {
        instrument {
          kubernetes {
            namespace = "demo"
          }
        }

        webhook {
          external_deployment_name = "beyla-k8s-injector/beyla-k8s-injector-controller-manager"
        }

        image_version = "2026.06.16.0"
        exporter_otlp_endpoint = "http://alloy-otlp:4318"
      }
```

Instead, I had to deploy my own alloy preconfigured to run `beyla.ebpf` with the injector mode. I guess it is expected as the current latest image of Alloy still does not support this Beyla feature, right?

PR #2766, opened 2026-07-01
Add EC2 tag enrichment data processor

### LINKED ISSUE #2678: [4.2] EC2 Enrichment Processor

petewall:
# Description

This adds the EC2 Data Processor, which can find tags on EC2 nodes and add them as labels or attributes to telemetry data.

For example, by setting this:

```yaml
dataProcessors:
  # Copies the EC2 instance `Team` tag onto telemetry, matching each node's `node` label to
  # its instance private DNS name via discovery.ec2.
  ec2-tags:
    type: ec2Enrichment
    region: ap-northeast-2
    tags:
      team: Team

clusterMetrics:
  enabled: true
  collector: alloy-metrics
  dataProcessors: [ec2-tags]
```

All cluster metrics with a `node` label will also get a `team` label from the `Team` tag on the matching EC2 instance.

So, given the EC2 instance `ip-10-0-1-2` has a `Team` tag of `central`, then a node-level metric
```
kube_node_info{cluster="my-cluster", node="ip-10-0-1-2", team="central"}
```

Fixes #2678

## Authorship


charts/k8s-monitoring/docs/examples/dataProcessors/custom/static-label/output.yaml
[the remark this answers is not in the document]
petewall: It's this:
```yaml
data:
  tenantId: "MQ=="
  username: "bG9raQ=="
  password: "bG9raXBhc3N3b3Jk"
```
Which corresponds to:
```
  tenantId: "1"
  username: loki
  password: lokipassword
```

Not an issue

charts/k8s-monitoring/dataProcessors/ec2Enrichment-values.yaml
TylerHelmuth, line 1: Since there is already a lot of docs here I think it is worth adding docs for precedence. For

```yaml
 example:
   tags:
     team: Team
```

lets document what happens if the telemetry flowing from feature to destination already has the label `team`

charts/k8s-monitoring/dataProcessors/ec2Enrichment-values.yaml
```diff
@@ -0,0 +1,49 @@
...
+region: ""
+
+# -- The ARN of the role to assume when discovering EC2 instances. If not set, the default
+# credentials from the AWS credential chain are used.
+# @section -- General
+roleARN: ""
+
+# -- How often to refresh the list of discovered EC2 instances and their tags.
+# @section -- General
+refreshInterval: ""
```
TylerHelmuth, line 44: Should we provide a reasonable default?

charts/k8s-monitoring/templates/dataProcessors/_dataProcessor_ec2Enrichment.tpl
```diff
@@ -0,0 +1,153 @@
+{{- define "secrets.list.ec2Enrichment-dataProcessor" }}{{ end -}}
```
TylerHelmuth: What does this empty template do?

charts/k8s-monitoring/templates/dataProcessors/_dataProcessor_ec2Enrichment.tpl
```diff
@@ -0,0 +1,153 @@
...
+{{- define "dataProcessors.ec2Enrichment.alloy.loki.logs.config" }}
+{{- include "dataProcessors.ec2Enrichment.alloy.enrichPipeline" (dict "processor" .processor "processorName" .processorName "type" "logs" "ecosystem" "loki" "enrichComponent" "loki.enrich" "matchLabelArg" "logs_match_label") }}
+{{- end -}}
+{{- define "dataProcessors.ec2Enrichment.alloy.pyroscope.profiles.config" }}
+{{- include "dataProcessors.ec2Enrichment.alloy.enrichPipeline" (dict "processor" .processor "processorName" .processorName "type" "profiles" "ecosystem" "pyroscope" "enrichComponent" "pyroscope.enrich" "matchLabelArg" "profiles_match_label") }}
+{{- end -}}
+
+{{- /* Sanitized telemetry label names to copy from the discovered instances. */}}
+{{- define "dataProcessors.ec2Enrichment.tagsToCopy" -}}
+{{- $tags := default dict (dig "tags" dict .) -}}
```
TylerHelmuth: NIT: I believe `{{- $tags := default dict .tags -}}` is equivalent. Both assume `.` is a map and `.tags` returns nil if `tags` is not in `.`. `default` treats both nil and empty map the same, using the fallback value.

charts/k8s-monitoring/docs/examples/dataProcessors/custom/scrub-log-pii/output.yaml
TylerHelmuth, line 1: Do we need to add an example specifically for this data processor?

charts/k8s-monitoring/dataProcessors/ec2Enrichment-values.yaml
petewall: Good point. I'll do that.

charts/k8s-monitoring/docs/examples/dataProcessors/custom/scrub-log-pii/output.yaml
petewall: Yeah, good call. I'll do that, too.

charts/k8s-monitoring/dataProcessors/ec2Enrichment-values.yaml
petewall: It allows you to override the default interval in Alloy:  docs/examples/features/host-metrics/linux-hosts-via-alloy/alloy-metrics.alloy. By having it blank, we use Alloy's default of `60s`, but we don't have to hardcode it, so if they change, we'll change, too.

What I can do is show the default value in the docs.

charts/k8s-monitoring/templates/dataProcessors/_dataProcessor_ec2Enrichment.tpl
petewall: That was making space for future secrets, but I'll remove it since it's unused as of now.

charts/k8s-monitoring/templates/dataProcessors/_dataProcessor_ec2Enrichment.tpl
petewall: True. much cleaner.

TylerHelmuth (approved):

PR #2767, opened 2026-07-01
K8s injection controller

mariomac:
# Description

Taking over https://github.com/grafana/k8s-monitoring-helm/pull/2760

Adds support for deploying the k8s sdk injection controller with appropriate RBAC permissions for Alloy to be able to write the config maps.

I made the change with couple of assumptions:
- Since we need to mark which service account is writing the config map for the injector, I pinned it to `grafana-k8s-monitoring-alloy-metrics`. I couldn't figure out how to retrieve the actual name.
- The Alloy PR where enabling the config map RBAC is still in review, so I decided to add custom RBAC for this purpose. I couldn't reuse the rbac.rules from the Alloy helm chart, because they don't seem to support setting namespace for the individual rule. Doing it that way would open up the new RBAC to the full cluster, which isn't desirable for the newly added RBAC.

## Authorship


charts/k8s-monitoring/values.yaml
```diff
@@ -501,6 +501,34 @@ telemetryServices:
...
+
+    # -- Render a Namespace object for `namespace.name`.
+    # @section -- SDK Injector
+    create: false
+
+  # -- Comma-separated allowlist of usernames permitted to create or update annotated injection ConfigMaps. By default,
+  # Kubernetes expands `$(POD_NAMESPACE)` to the namespace where the SDK Injector pod is running.
+  # We add grafana-k8s-monitoring-alloy-metrics as the service account allowed to write the config maps.
+  # @section -- SDK Injector
+  allowedConfigMapWriters: "system:serviceaccount:$(POD_NAMESPACE):grafana-k8s-monitoring-alloy-metrics"
```
mariomac: Linking comment https://github.com/grafana/k8s-monitoring-helm/pull/2760#discussion_r3504750420

charts/k8s-monitoring/templates/_sdk_injector_extra_objects.tpl
```diff
@@ -0,0 +1,53 @@
...
+Do not use Alloy `spec.rbac.rules` for this temporary permission. Without
+`spec.rbac.namespaces`, Alloy renders those rules as cluster-scoped RBAC; setting
+`spec.rbac.namespaces` would scope every Alloy RBAC rule, not just this SDK
+Injector ConfigMap writer rule. Rendering a separate Role/RoleBinding keeps only
+this permission scoped to the namespace where the Alloy collector is deployed.
+*/}}
+{{- define "sdkInjector.extraObjects" -}}
+{{- $sdkInjector := index .Values "sdkInjector" | default dict }}
+{{- if $sdkInjector.enabled }}
+{{- range $collectorName := include "collectors.list.enabled" . | fromYamlArray }}
```
mariomac: Linking comment https://github.com/grafana/k8s-monitoring-helm/pull/2760#discussion_r3504810351

charts/k8s-monitoring/values.yaml
```diff
@@ -501,6 +501,34 @@ telemetryServices:
...
+# @default -- Disabled
+# @section -- SDK Injector
+sdkInjector:
+  # -- Deploy the SDK Injector and render namespace-scoped RoleBindings granting Alloy collectors permission to write injection ConfigMaps.
+  # @section -- SDK Injector
+  enabled: false
+
+  # -- Set a proper name to avoid RELEASE-NAME-sdkInjector
+  # @ignored
+  nameOverride: k8s-injection-controller
```
petewall, line 515: Do we need this? Wouldn't this conflict if we had multiple deployments?

charts/k8s-monitoring/values.yaml
```diff
@@ -501,6 +501,34 @@ telemetryServices:
...
+
+  namespace:
+    # -- Namespace where the SDK Injector is installed. Leave empty to use the Helm release namespace.
+    # @section -- SDK Injector
+    name: ""
+
+    # -- Render a Namespace object for `namespace.name`.
+    # @section -- SDK Injector
+    create: false
```
petewall, line 524: I'd probably leave these off.

charts/k8s-monitoring/values.yaml
petewall: This really concerns me, as it hardcodes the release name and the collector name at the same time.
I'm guessing this is somethign that has to be defined in the values file here, right.

petewall (commented):
I'd like to move this into telemetryServices, too

petewall:
Closing in favor of https://github.com/grafana/k8s-monitoring-helm/pull/2765

PR #2779, opened 2026-07-05
Set memory request on alloy container

philipmw:
Without this, the pod's overall memory request is only 50 MiB, because only `config-reloader` sets a memory request value. This causes the pod to always appear to exceed its memory request.

Max memory usage of `alloy` container observed over 5 days: 116 MiB.

Verified by updating my own values.yaml:

```
      ~ values                     = [
          ~ <<-EOT
                ...
                collectors:
                  alloy-metrics:
              +     alloy:
              +       resources:
              +         requests:
              +           memory: "128Mi"
                    presets:
                      - clustered
                      - statefulset
                  alloy-singleton:
              +     alloy:
              +       resources:
              +         requests:
              +           memory: "128Mi"
                    presets:
                      - singleton
                  alloy-logs:
              +     alloy:
              +       resources:
              +         requests:
              +           memory: "128Mi"
                    presets:
                      - filesystem-log-reader
                      - daemonset
```

And observed that the pod no longer appears to exceed its memory request, because both containers set an appropriate memory request.

Screenshot_20260705_131044

## Authorship

### DIFF
--- a/charts/k8s-monitoring/collectors/upstream/alloy-values.yaml
+++ b/charts/k8s-monitoring/collectors/upstream/alloy-values.yaml
@@ -95,7 +95,9 @@ alloy:
   # -- Security context to apply to the Grafana Alloy container.
   securityContext: {}
   # -- Resource requests and limits to apply to the Grafana Alloy container.
-  resources: {}
+  resources:
+    requests:
+      memory: "128Mi"
   # -- Set lifecycle hooks for the Grafana Alloy container.
   lifecycle: {}
   # preStop:


TylerHelmuth (commented):
It is hard for us to guess a good default for the collectors because it is based on the configuration and the amount of data flowing through. For this reason the upstream collector chart actually removed its presets (https://github.com/open-telemetry/opentelemetry-helm-charts/issues/644).

[Instead, checkout the collector presets that are an easy way to set resources for the install collectors. ](https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring/docs/collectors/presets)

PR #2784, opened 2026-07-07
fix: prefer explicit service_name label over job-derived service.name in OTLP destinations

### LINKED ISSUE #2783: autoInstrumentation + OTLP-only destination mangles Beyla service identity
## Summary

When `autoInstrumentation` (Beyla) is enabled and metrics are shipped through an **OTLP-only destination**, every Beyla-discovered service gets a mangled identity: the stored `job` label becomes `<namespace>/<namespace>/<workload>` and the derived service name becomes `<namespace>/<workload>` instead of `<workload>`.

Consequences:

- Application Observability shows Beyla services named `<namespace>/<workload>`.
- The **Logs tab for those services is always blank**: the plugin queries Loki with `{service_namespace="<ns>", service_name="<ns>/<workload>"}`, while `podLogsViaOpenTelemetry` labels the same pods' logs with the plain `service_name="<workload>"`. The selector never matches.
- `target_info` carries merged conflicting attributes, e.g. `service_name="<ns>/<workload>;<workload>"` and `k8s_cluster_name="<cluster>;<cluster>"`.
- Services instrumented with OTel SDKs (traces via `applicationObservability`) get clean identities, so each workload can appear twice in the service list (one working, one broken).

Deployments that also have a `prometheus`-type destination are unaffected for the remote_write path, which masks the bug in mixed setups.

## Environment

- Chart: `k8s-monitoring` 4.2.0 (also reproduced on 4.1.4)
- Beyla: 3.14.0 and 3.20.0 (version-independent)
- Destination: single `type: otlp` destination with `metrics.enabled: true`
- Features: `autoInstrumentation` with `prometheus_export.features: [application, application_span, application_service_graph, network]`, `podLogsViaOpenTelemetry`, `applicationObservability`

## Reproduction

1. Install `k8s-monitoring` with an **OTLP destination only** and the features above.
2. Deploy any HTTP workload (we used `open-telemetry/opentelemetry-demo` 0.40.9 with its built-in collector disabled).
3. Wait for Beyla metrics, then query:

```promql
count by (job) (traces_spanmetrics_calls_total{job=~".+/.+/.+"})
```

Result (should be empty; every entry is a Beyla-discovered service):

```
{job="otel-demo/otel-demo/cart"}
{job="otel-demo/otel-demo/frontend"}
{job="otel-demo/otel-demo/load-generator"}
...
```

4. In Application Observability, open one of these services and the Logs tab: the generated query `{service_namespace="otel-demo", service_name="otel-demo/cart"}` returns no streams, while the logs exist in Loki with `service_name="cart"`.

## Root cause

Chain across three places:

1. Beyla's `prometheus_export` emits scrape-side `job="<ns>/<workload>"` plus plain `service_name`/`service_namespace` labels.
2. `charts/feature-auto-instrumentation/templates/_module.alloy.tpl`: the `prometheus.scrape "beyla_applications"` component uses `honor_labels = true` and forwards to the destination with no `job` relabeling.
3. `templates/destinations/_destination_otlp.tpl`: metrics enter `otelcol.receiver.prometheus`, which per OTel spec sets resource `service.name` to the **entire** `job` string (`"<ns>/<workload>"`). The destination's transform processor only promotes the datapoint `service_name` attribute to the resource `where resource.attributes["service.name"] == nil or == ""` - it is already set, so the correct plain value never wins. `service_namespace` (empty at that point) does get promoted.
4. The Grafana Cloud OTLP gateway then rebuilds `job = service.namespace + "/" + service.name` = `<ns>/<ns>/<workload>`.

mbaykara:
Fixes #2783

When metrics from a Prometheus scrape (e.g. Beyla's `prometheus_export`) are sent through an OTLP destination, `otelcol.receiver.prometheus` sets resource `service.name` to the full `job` label - for Beyla that is `<namespace>/<workload>`. The existing transform only promotes the explicit `service_name` datapoint attribute when the resource attribute is empty, so the wrong value survives. The OTLP gateway then reconstructs `job` as `<ns>/<ns>/<workload>`, Application Observability derives `service_name="<ns>/<workload>"`, and its Logs tab queries match nothing (pod logs carry the plain `service_name="<workload>"`). It also produces the `;`-joined conflict values in `target_info`.

This adds one transform statement: overwrite resource `service.name` with the explicit `service_name` datapoint attribute when the resource value is job-shaped (contains `/`) and differs. OTLP-native telemetry is unaffected (no `service_name` datapoint attribute); prometheus destinations are unaffected (template not involved).

Verified end-to-end on a live cluster (chart 4.2.0, Beyla, OTLP-only destination, OTel demo app): stored series become `job="otel-demo/cart"` / `service_name="cart"`, the Application Observability Logs tab populates, and the duplicate `x;y` values disappear.

### DIFF
--- a/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
+++ b/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
@@ -18,9 +18,38 @@
 {{- if eq (include "destinations.otlp.supports_metrics" .) "true" }}
 otelcol.receiver.prometheus {{ include "helper.alloy_name" $.destinationName | quote }} {
   output {
-    metrics = [{{ include "destinations.otlp.alloy.otlp.metrics.target" (dict "destination" . "destinationName" $.destinationName) | trim }}]
+    metrics = [otelcol.processor.transform.{{ include "helper.alloy_name" $.destinationName }}_scrape_normalize.input]
   }
 } // otelcol.receiver.prometheus "{{ include "helper.alloy_name" $.destinationName }}"
+
+// Repairs artifacts of the Prometheus-to-OTLP conversion performed by
+// otelcol.receiver.prometheus. Only metrics that came from a Prometheus scrape pass
+// through this component; OTLP-native telemetry enters the pipeline further down.
+otelcol.processor.transform {{ printf "%s_scrape_normalize" (include "helper.alloy_name" $.destinationName) | quote }} {
+  error_mode = {{ .processors.transform.errorMode | quote }}
+  metric_statements {
+    context = "resource"
+    statements = [
+      // The conversion sets service.name to the full Prometheus job label (e.g.
+      // "namespace/workload"). When the scrape target declares its identity explicitly
+      // with a service_name label on its own target_info metric (e.g. Grafana Beyla),
+      // prefer that value and drop the duplicate.
+      `set(attributes["service.name"], attributes["service_name"]) where attributes["service_name"] != nil and attributes["service_name"] != ""`,
+      `delete_key(attributes, "service_name") where attributes["service_name"] == attributes["service.name"]`,
+    ]
+  }
+  metric_statements {
+    context = "datapoint"
+    statements = [
+      // Same correction for scrape targets that put service_name on every series
+      // instead of exposing a target_info metric.
+      `set(resource.attributes["service.name"], attributes["service_name"]) where attributes["service_name"] != nil and attributes["service_name"] != ""`,
+    ]
+  }
+  output {
+    metrics = [{{ include "destinations.otlp.alloy.otlp.metrics.target" (dict "destination" . "destinationName" $.destinationName) | trim }}]
+  }
+} // otelcol.processor.transform "{{ include "helper.alloy_name" $.destinationName }}_scrape_normalize"
 {{- end }}
 {{- if eq (include "destinations.otlp.supports_logs" .) "true" }}
 otelcol.receiver.loki {{ include "helper.alloy_name" $.destinationName | quote }} {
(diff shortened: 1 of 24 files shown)


charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -120,6 +120,12 @@ otelcol.processor.transform {{ include "helper.alloy_name" $.destinationName | q
 {{- range $datapointAttribute, $resourceAttribute := .processors.transform.metrics.datapointToResource }}
   {{- if $resourceAttribute }}
 {{- if or (eq $resourceAttribute "service.name") (eq $resourceAttribute "service.namespace") }}
+{{- if eq $resourceAttribute "service.name" }}
+      {{- /* Metrics converted from a Prometheus scrape get service.name set to the full job
+      label (e.g. "namespace/workload" from Beyla). When the scrape also carried an explicit
+      service_name label, prefer it so the identity matches logs and traces. */}}
+      `set(resource.attributes[{{ $resourceAttribute | quote }}], attributes[{{ $datapointAttribute | quote }}] ) where attributes[{{ $datapointAttribute | quote }}] != nil and attributes[{{ $datapointAttribute | quote }}] != "" and resource.attributes[{{ $resourceAttribute | quote }}] != nil and resource.attributes[{{ $resourceAttribute | quote }}] != attributes[{{ $datapointAttribute | quote }}] and IsMatch(resource.attributes[{{ $resourceAttribute | quote }}], "^.+/.+$")`,
```
TylerHelmuth: I am a bit worried (because of the regex check) that this is a breaking change. Are we sure it is always safe to overwrite any `service.name` resource attribute that has a `"^.+/.+$"` pattern? I am worried this could accidentally break some users who are setting a service.name explicitly.

mbaykara:
Landed via #2788 (both fixes were merged together there). Closing.

PR #2788, opened 2026-07-07
fix: reconcile scraped target_info resource attributes in OTLP destinations

### LINKED ISSUE #2786: OTLP destination: scraped target_info resource attributes conflict with chart-set dotted attributes, producing ';'-joined label values
## Summary

Follow-up to #2783 / #2784. With an OTLP destination and Beyla (`autoInstrumentation`) enabled, `target_info` ends up with duplicate `;`-joined values:

```
target_info{..., k8s_cluster_name="my-cluster;my-cluster", service_name="cart;cart", service_namespace="otel-demo;otel-demo"}
```

The `;` is the Prometheus OTLP endpoint's conflict resolution: two different resource attributes translate to the same Prometheus label name.

## Root cause (verified via `otelcol.exporter.debug`)

Beyla's Prometheus endpoint exposes its own `target_info` metric with flat labels. Per the OTel <-> Prometheus compatibility spec, `otelcol.receiver.prometheus` converts a scraped `target_info` back into resource attributes **verbatim**:

```
Resource attributes:
     -> k8s_cluster_name: Str(aks-monitoring)   # underscore, from scraped target_info
     -> service_name: Str(cart)
     -> service_namespace: Str(otel-demo)
```

The destination's transform then adds the dotted variants (`k8s.cluster.name`, promoted `service.namespace`) at resource level. Both variants map to the same Prometheus label, so the endpoint joins the values.

## Impact

Garbage `target_info` labels; anything joining on it (Asserts, Entity Graph, Application Observability enrichment) is affected.

mbaykara:
Fixes #2786. Includes the reworked #2783 fix as its base commit (see below) - if the direction here works, #2784 will be updated to match.

## Problem

Scrape targets that expose their own `target_info` metric (e.g. Grafana Beyla) get its labels converted verbatim into flat resource attributes (`service_name`, `service_namespace`, `deployment_environment`, `k8s_cluster_name`) by `otelcol.receiver.prometheus`. The destination transform then adds the dotted equivalents, and since both variants translate to the same Prometheus label, the OTLP endpoint joins the values with `;` (`k8s_cluster_name="c;c"`, `service_name="cart;cart"`), polluting `target_info` and everything that joins on it (Asserts, Entity Graph, Application Observability enrichment).

## Approach

Following the review feedback on #2784 (a value-pattern check on `service.name` could break users who set such names on purpose), the corrections no longer guess provenance from the data. A dedicated `otelcol.processor.transform "<destination>_scrape_normalize"` now sits directly between `otelcol.receiver.prometheus` and the rest of the destination pipeline. Everything passing through it came from a Prometheus scrape, so:

- `service.name` synthesized from the `job` label is replaced with the explicit `service_name` label when the target provides one - no regex heuristics needed.
- The flat `service_namespace`, `deployment_environment`, and `deployment_environment_name` resource attributes are promoted to their dotted counterparts and the flat copies dropped when equal.
- OTLP-native telemetry never enters this component, so user-set resource attributes cannot be affected.

The only statement left in the shared transform is the `k8s_cluster_name` dedup, because `k8s.cluster.name` is set from chart values there (after the normalize step). It is an equality-guarded `delete_key`, so it only removes an exact duplicate.

The statements are grouped and commented in the rendered config.

## Verification

Verified on a live cluster (chart 4.2.0 + Beyla + OTLP-only destination, equivalent statements applied via destination `processors.transform` values): `target_info` values become single (`service_name="cart"`, no `;`), the Application Observability service identity matches the pod logs, and its Logs tab populates.

### DIFF
--- a/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
+++ b/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
@@ -18,9 +18,38 @@
 {{- if eq (include "destinations.otlp.supports_metrics" .) "true" }}
 otelcol.receiver.prometheus {{ include "helper.alloy_name" $.destinationName | quote }} {
   output {
-    metrics = [{{ include "destinations.otlp.alloy.otlp.metrics.target" (dict "destination" . "destinationName" $.destinationName) | trim }}]
+    metrics = [otelcol.processor.transform.{{ include "helper.alloy_name" $.destinationName }}_scrape_normalize.input]
   }
 } // otelcol.receiver.prometheus "{{ include "helper.alloy_name" $.destinationName }}"
+
+// Repairs artifacts of the Prometheus-to-OTLP conversion performed by
+// otelcol.receiver.prometheus. Only metrics that came from a Prometheus scrape pass
+// through this component; OTLP-native telemetry enters the pipeline further down.
+otelcol.processor.transform {{ printf "%s_scrape_normalize" (include "helper.alloy_name" $.destinationName) | quote }} {
+  error_mode = {{ .processors.transform.errorMode | quote }}
+  metric_statements {
+    context = "resource"
+    statements = [
+      // The conversion sets service.name to the full Prometheus job label (e.g.
+      // "namespace/workload"). When the scrape target declares its identity explicitly
+      // with a service_name label on its own target_info metric (e.g. Grafana Beyla),
+      // prefer that value and drop the duplicate.
+      `set(attributes["service.name"], attributes["service_name"]) where attributes["service_name"] != nil and attributes["service_name"] != ""`,
+      `delete_key(attributes, "service_name") where attributes["service_name"] == attributes["service.name"]`,
+    ]
+  }
+  metric_statements {
+    context = "datapoint"
+    statements = [
+      // Same correction for scrape targets that put service_name on every series
+      // instead of exposing a target_info metric.
+      `set(resource.attributes["service.name"], attributes["service_name"]) where attributes["service_name"] != nil and attributes["service_name"] != ""`,
+    ]
+  }
+  output {
+    metrics = [{{ include "destinations.otlp.alloy.otlp.metrics.target" (dict "destination" . "destinationName" $.destinationName) | trim }}]
+  }
+} // otelcol.processor.transform "{{ include "helper.alloy_name" $.destinationName }}_scrape_normalize"
 {{- end }}
 {{- if eq (include "destinations.otlp.supports_logs" .) "true" }}
 otelcol.receiver.loki {{ include "helper.alloy_name" $.destinationName | quote }} {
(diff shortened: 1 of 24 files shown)


petewall (changes requested):
Thanks for the change and I applaud how you discovered the root cause!
A few changes I'd like to see:
1. Add a line in a new "## Unreleased" section in the CHANGELOG.md file.
2. Add `deployment_environment` / `deployment.environment` to the list of attributes to handle.

mbaykara:
Thanks for the review! Both addressed in 728ba66:

1. Added a `## Unreleased` section to CHANGELOG.md with an entry for this fix.
2. Added reconciliation pairs for `deployment_environment` -> `deployment.environment` and `deployment_environment_name` -> `deployment.environment.name` (both variants, matching the existing `datapointToResource` defaults).

charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -84,6 +84,20 @@ otelcol.processor.transform {{ include "helper.alloy_name" $.destinationName | q
 {{- range $resourceAttribute := $resourceAttributesToRemove }}
       `delete_key(attributes, {{ $resourceAttribute | quote }})`,
 {{- end }}
+      {{- /* Metrics converted from a Prometheus scrape can carry flat resource attributes
+      (service_name, service_namespace, deployment_environment, k8s_cluster_name) when the
+      target exposes its own target_info metric (e.g. Beyla). These conflict with the dotted
+      attributes set above and get ";"-joined by the Prometheus OTLP endpoint. Reconcile them
+      into the dotted attributes and drop the duplicates. */}}
+      `delete_key(attributes, "k8s_cluster_name") where attributes["k8s_cluster_name"] == attributes["k8s.cluster.name"]`,
```
TylerHelmuth, line 129: IDK if we have precedence for this but these statements are complex enough that I think it would be nice if we rendered them with some grouping and config comments explaining to the user what they do.

charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -84,6 +84,20 @@ otelcol.processor.transform {{ include "helper.alloy_name" $.destinationName | q
...
+      `set(attributes["service.namespace"], attributes["service_namespace"]) where (attributes["service.namespace"] == nil or attributes["service.namespace"] == "") and attributes["service_namespace"] != nil and attributes["service_namespace"] != ""`,
+      `delete_key(attributes, "service_namespace") where attributes["service_namespace"] == attributes["service.namespace"]`,
+      `set(attributes["service.name"], attributes["service_name"]) where attributes["service_name"] != nil and attributes["service_name"] != "" and (attributes["service.name"] == nil or attributes["service.name"] == "" or (attributes["service.name"] != attributes["service_name"] and IsMatch(attributes["service.name"], "^.+/.+$")))`,
```
TylerHelmuth: I have the same concern here as in #2784. Are we sure it is ok to override a set `service.name` what i has this pattern?

mbaykara:
You're right, the regex check was guessing intent from the value, and a slash in a user-set service.name is perfectly legal. Restructured along the same lines as discussed in #2784: the reconciliation statements now live in the dedicated `*_scrape_normalize` transform between `otelcol.receiver.prometheus` and the rest of the destination pipeline, grouped and with comments in the rendered config explaining what each block does. That also answers the `service.name` concern here  only scrape-converted metrics pass through that component. One exception: the `k8s_cluster_name` dedup stays in the shared transform, because the chart only sets `k8s.cluster.name` there (after the normalize step). It's an equality-guarded `delete_key`, so it only removes an exact duplicate and never overwrites anything.

This branch currently carries the reworked #2783 fix as its base commit so the whole approach is reviewable in one place
 if the direction works for you I'll update #2784 to match.

TylerHelmuth (approved):

petewall (approved):

PR #2790, opened 2026-07-07
Add filesystem-log-reader preset to iHub test

### LINKED ISSUE #2773: Clean up instrumentation-hub presets
I'm taking a look at the current [preset values](https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring/tests/platform/grafana-cloud/instrumentation-hub/preset-values) we're testing against for the instrumentation hub, and would like to propose a couple of changes:

* Remove `demo.yaml`. I thought we'd discussed that we will not be pursuing this. Any reason to keeping it around? Otherwise it's just more things to keep a mental note of
* Rename `my-collector` templates. When deploying these, they felt a little toy-like to me. Can we find a better name?

### LINKED ISSUE #2789: Instrumentation hub presets do not allow for logs collection
The previous iteration of the instrumentation hub used the [grafana-cloud-onboarding-helm-chart](https://github.com/grafana/grafana-cloud-onboarding-helm-chart).  As we're folding it into the k8s monitoring chart, the assumption was that pipelines would work the same way.

Unfortunately, the [presets](https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring/tests/platform/grafana-cloud/instrumentation-hub/preset-values) for instrumentation hub don't seem to mount /var/log/pods (or equivalent) so logs collection fails.

Depending on whether this is an internal Chart-only change or also needs updated instructions, this might also need a FE change to point to updated presets.

petewall:
# Description

For pod logs to work from Instrumentations Hub, we need to add the `filesystem-log-reader` preset to the daemonset collector.

We're also going to remove the demo example values and stop using the collector name 'my-collector'.

Fixes #2789 
Fixes #2773

## Authorship


charts/k8s-monitoring/tests/platform/grafana-cloud/instrumentation-hub/preset-values/minimal.yaml
```diff
@@ -13,8 +13,8 @@ collectorCommon:
         password: my-remote-cfg-password
 
 collectors:
-  my-collector:
-    presets: [large, privileged, clustered, service-discovery, daemonset]
+  alloy:
```
TylerHelmuth, line 16: This collector sets `daemonset` preset but is named `alloy` instead of `alloy-daemonset` like `maximum` and `standard` values. Is that ok?

TylerHelmuth (approved):

charts/k8s-monitoring/tests/platform/grafana-cloud/instrumentation-hub/preset-values/minimal.yaml
petewall: Yep. The name is independent of the presets, and I actually wanted to use `alloy` instead of `alloy-daemonset` to highlight that.
This one only has one alloy collector, while the others have two, with their main distinction being one is a daemonset and the other is a deployment.

PR #2793, opened 2026-07-07
Explicitly set `false` to Beyla's K8s Cache and SDK Injector Controller

### LINKED ISSUE #2781: After upgrading to 4.2.0 k8s-injection-controller gets OOMKilled with presets
After upgrading to 4.2.0 k8s-injector-controller get OOMKilled with the current chart presets. While an easy fix I would recommend that the default limits get raised.

### LINKED ISSUE #2782: `k8s-injection-controller` gets deployed by default
Rendering the Helm chart `v4.2.0` (using `helm template`) with the following values:

```yaml
cluster:
  name: my-cluster

destinations:
  loki:
    type: loki
    url: http://loki-gateway.loki.svc.cluster.local/loki/api/v1/push

clusterEvents:
  enabled: true
  collector: alloy-logs

podLogsViaKubernetesApi:
  enabled: true

selfReporting:
  enabled: false

collectors:
  alloy-logs:
    enabled: true
    alloy:
      resources:
        requests:
          cpu: 50m
          memory: 128Mi
        limits:
          cpu: 1
          memory: 1Gi
      enableReporting: false
      mounts:
        varlog: false
        dockercontainers: false
      clustering:
        enabled: true
```

I get all the `k8s-injection-controller`/`sdkInjector` deployments, RBAC, etc even though they are disabled by default in the `telemetry-services` sub-chart with `sdkInjector.deploy = false`.

Workaround is to explicitly add the following to `values.yaml`:

```yaml
telemetryServices:
  sdkInjector:
    deploy: false
```

EDIT: I noticed now that the same workaround is needed for Beyla by setting `beyla.deploy = false`.

petewall:
# Description

Explicitly set `false` for the new deployments in 4.2: Beyla’s K8s Cache and the SDK Injector Controller
This also gives nice targets for users when they want to enable them.

Fixes #2781 (because the sdk injector shouldn't have been deployed)
Fixes #2782

## Authorship

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+*   Surface `deploy: false` defaults for `telemetryServices.beyla` and `telemetryServices.sdkInjector` in the top-level values, matching the other telemetry services. This prevents them from deploying unintentionally on Helm versions affected by a subchart value coalescing regression (helm/helm#32132). (#2781, #2782) (@petewall)
 *   Change the default remote configuration `pollFrequency` from `5m` to `30s`. (@TylerHelmuth)
 *   Add `CLUSTER_NAME` environment variable to remote-config-enabled collectors alongside `NAMESPACE` and `POD_NAME`. (#2775) (@petewall)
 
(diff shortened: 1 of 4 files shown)


TylerHelmuth (approved):
damn coalesce bug

PR #2794, opened 2026-07-07
fix: restore openshift platform test

TylerHelmuth:
# Description

Restores the openshift platform test

## Authorship

### DIFF
--- a/charts/k8s-monitoring/tests/platform/openshift/test-plan.yaml
+++ b/charts/k8s-monitoring/tests/platform/openshift/test-plan.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: helm-chart-toolbox.grafana.com/v1
 kind: TestPlan
-name: gke
+name: openshift
 
 subject:
   releaseName: k8smon
--- a/.github/workflows/platform-test.yml
+++ b/.github/workflows/platform-test.yml
@@ -81,20 +81,23 @@ jobs:
           repository: grafana/helm-chart-toolbox
           persist-credentials: 'false'
 
-      - name: Set up Helm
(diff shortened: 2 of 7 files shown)


TylerHelmuth:
To make the test less flaky I want to increase the timeout. See https://github.com/grafana/helm-chart-toolbox/pull/122

charts/k8s-monitoring/tests/platform/openshift/test-plan.yaml
petewall, line 12: Rather than requiring the change to add a timeout env var to helm-chart-toolbox, we could set the extra helm args here:
```suggestion
subject:
  releaseName: k8smon
  path: ../../..
  valuesFile: values.yaml
  set:
    - key: "cluster.name"
      valueFrom: clusterName
  extraArgs: "--timeout 10m"
```

petewall (commented):
Is there no way to use the helm-chart-toolbox for cluster creation? We could amend the code inside of https://github.com/grafana/helm-chart-toolbox/blob/main/tools/helm-test/includes/cluster/openshift.sh to get the new rules.

charts/k8s-monitoring/tests/platform/openshift/test-plan.yaml
TylerHelmuth: I am pretty sure `extraArgs` doesn't get used in the `helm test` command, which is the command I want to have a longer timeout (the default is 5m and sometimes that isn't enough)

petewall: Ah, gotcha. Yeah, it's a helm test arg, not a helm install arg...

PR #2799, opened 2026-07-08
fix: update privilege preset to have more permission to enable IHub

### LINKED ISSUE #2796: instrumentation hub presets don't enable hostNetwork
Opening a new issue for another small drift between the new presets for instrumentation hub and the old grafana-cloud-onboarding-helm chart.

It looks like the current set of instructions does not set `hostNetwork` to true, which is required for Beyla's injector to function properly. Looks like the current presets also set a different `dnsPolicy`. The old chart used to set `ClusterFirstWithHostNet` while the current one does `ClusterFirst`.

Sounds like that in the `privileged` preset we only set up hostPID
https://github.com/grafana/k8s-monitoring-helm/blob/070bb2e58b60c63fd3863776f4e6fde59a043b48/charts/k8s-monitoring/docs/collectors/presets/privileged.md?plain=1#L38-L39

Whereas in the past we also set up hostNetwork and the dnsPolicy explicitly

https://github.com/grafana/grafana-cloud-onboarding-helm-chart/blob/cd7c12984b3dc6d2146f93f72a2e447e7b943c00/charts/grafana-cloud-onboarding/values.yaml#L51-L65

### LINKED ISSUE #2797: instrumentation hub presets differ in mounted volumes
The previous iteration of the instrumentation hub used the [grafana-cloud-onboarding-helm-chart](https://github.com/grafana/grafana-cloud-onboarding-helm-chart). 

Another difference we've noticed with the current instrumentation hub presets is in mounted volumes.

Previously, we had some host volumes like alloy-storage(hostPath, writable) + cgroup. Following the new instructions, we get no mounts.

Similarly, the Alloy storage path is now `/tmp/alloy` which differs from the previous instructions. Were these conscious design decisions or possible gaps?

The Alloy one is more straightforward to understand tradeoffs about, but I'm not sure how the cgroup thing, and maybe we should consult with Beyla folks to understand if/how they might use it?

TylerHelmuth:
# Description

Adds new, specific presets for root security context, host network, cgroups, and host storage. This allows easier mixing and matching of different elevated permissions for different types of installs.

Also update openshift SecurityContextConstraints to check for these settings like it does for host ports.

- Fixes https://github.com/grafana/k8s-monitoring-helm/issues/2796
- Fixes https://github.com/grafana/k8s-monitoring-helm/issues/2797

## Authorship


petewall:
Let's create new presets for these.
* `privileged` gets `securityContext.privileged: true`
* `host-storage`
  * Adds the storagePath, extra volume and extra mount.
* `host-network`
  * Adds hostPID, host, and dnsPolicy
* `host-cgroup`
  * Adds the cgroup extra volume and extra mount

petewall:
Doing all of that will also address https://github.com/grafana/k8s-monitoring-helm/issues/2797

petewall (approved):

PR #2800, opened 2026-07-08
Allow for destinations to be disabled

### LINKED ISSUE #2795: Allow `destinations.{dest}: null` in values schema
Hello, thank you for maintaining this great chart.

We leverage multiple values files as a way avoid repetition across environments: `-f common.yaml -f dev.yaml` -- dev merges over common, etc.

We used to be able to define all destinations in `common.yaml` then selectively disable them per-environment by nulling them out. This was a very useful configuration method because especially loki and otlp destination configs can get quite large and complicated with loki processing stages and ottl rules. Our common config deploys receiver and samplers and servicegraph generators meanwhile some environments dont have OTel at all so these workloads are unneeded, being able to exclude them with a single `null` was useful.

It looks like a recent schema change now prevents nulling destination objects.

fwiw, a `destinations.{dest}.enabled: (default true)` value would work just as well.

Thank you for your consideration.

petewall:
# Description

Allow for destinations to set `disabled=true` to remove them from being used.

Fixes #2795

## Authorship

### DIFF
--- a/charts/k8s-monitoring/schema-mods/destination.json
+++ b/charts/k8s-monitoring/schema-mods/destination.json
@@ -21,7 +21,10 @@
     "destination-map": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "#/definitions/destination"
+        "anyOf": [
+          { "type": "null" },
+          { "$ref": "#/definitions/destination" }
+        ]
       }
     },
     "custom-destination": {"properties": {"type": {"type": "string", "const": "custom"}}},
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+*   Allow for destinations to be removed by setting them to `null` (#2800) (@petewall)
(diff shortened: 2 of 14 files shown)


charts/k8s-monitoring/schema-mods/destination.json
```diff
@@ -21,7 +21,10 @@
     "destination-map": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "#/definitions/destination"
+        "anyOf": [
+          { "type": "null" },
```
TylerHelmuth, line 25: I am a bit worried we are going to run into our own `null` coalesce bugs if we go this route. I'd vote for the more explicit `disabled: true` approach so that we don't get any weird subcharting coalesce issues.

petewall: That's fair. Helm has not proven the most reliable when it comes with overriding values with nulls or `false`'s

I can adjust this to permit a destination to work like:
```yaml
destinations:
  myDestination:
    ...
    disabled: true
```

or `--set destinations.myDestination.disabled=true`

TylerHelmuth: I think that'll be safer lol also hopefully easier for readers/robots to understand.

petewall:
Merged in #2843

PR #2811, opened 2026-07-09
fix: beyla.service.targetPort now correctly sets other ports as documented

### LINKED ISSUE #2114: Beyla port needs to be set three times
If you want to override the beyla port, you have to set it three times:

```
autoInstrumentation:
  enabled: true
  beyla:
    service:
      targetPort: 9095

    config:
      data:
        internal_metrics:
          prometheus:
            port: 9095
        prometheus_export:
          port: 9095
```

But the docs say that just changing `beyla.service.targetPort` should work.

TylerHelmuth:
# Description

Fixes a bug where we were merging in the wrong direction resulting in unexpected behavior

- Fixes https://github.com/grafana/k8s-monitoring-helm/issues/2114

## Authorship

### DIFF
--- a/charts/k8s-monitoring/charts/feature-auto-instrumentation/templates/_beyla-config.tpl
+++ b/charts/k8s-monitoring/charts/feature-auto-instrumentation/templates/_beyla-config.tpl
@@ -39,5 +39,5 @@
   {{- end }}
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


petewall (commented):
... that's all it needed?

petewall (approved):

PR #2823, opened 2026-07-10
Add tracefs to instrumentation hub deployment

grcevski:
# Description

This is a port of the corresponding Beyla helm chart PR https://github.com/grafana/beyla/pull/2960. 

TL;DR

There's currently a kernel bug that has been backported all the way to 6.6 minor that is affecting any eBPF program that is using socket maps. We use socket maps for the traceparent context propagation logic and we were affected by it. There's a mitigation code introduced in OBI (therefore in Beyla) that works around the kernel bug, but we needed to add two tracepoints. For those we need tracefs.

There's work in progress on OBI side to rewrite the tracepoints to raw tracepoints, but until that happens we need access to tracefs. 

This PR is not urgent, but once Beyla with the bug mitigation code is added to Alloy, we'll need this change too in order to complete the fix.

## Authorship


charts/k8s-monitoring/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 *   Fix `unrecognized block name "rule_namespace"` error when using the `rules.namespaces` field of the `prometheus` destination to limit the PrometheusRules object discovery to specific namespaces (#2802) (@sebasnabas)
 *   Fix `autoInstrumentation.beyla.service.targetPort` not propagating to the generated Beyla config ports, which previously required setting the port in three places. (#2811) (@TylerHelmuth)
 *   Update Beyla to 1.16.10 (@TylerHelmuth)
+*   Add `host-tracefs` collector preset where `host-cgroup` is used.
```
TylerHelmuth: I dont see a connection in the PR between `host-tracefs` and `host-cgroup`. Are they always supposed to be used together?

charts/k8s-monitoring/Makefile
```diff
@@ -7,6 +7,13 @@ endif
 
 HAS_HELM_UNITTEST := $(shell $(HELM) plugin list | grep unittest 2> /dev/null)
 HAS_SHELLSPEC := $(shell command -v shellspec;)
+HAS_YQ := $(shell command -v yq 2>/dev/null)
+
+ifeq ($(HAS_YQ),)
+YQ = docker run --platform linux/amd64 --rm --volume $(shell pwd):/workdir -w /workdir mikefarah/yq
```
TylerHelmuth: We are looking to move some of our local commands to docker but I'd prefer to keep that chore separate from feature adds.

charts/k8s-monitoring/CHANGELOG.md
grcevski: They are essentially not related, but they are both needed for the same feature. If you enable context-propagation for OBI/Beyla, you need to mount the cgroup so we can attach the BPF program that does the injection of the headers, and now we need tracefs to mitigate this kernel bug.

charts/k8s-monitoring/Makefile
grcevski: I had some issues with running yq locally, then I installed it. I thought I removed this.  Sorry about it, I'll revert.

grcevski: reverted.

charts/k8s-monitoring/CHANGELOG.md
TylerHelmuth: Ok cool then I think we can simplify the changelog since they are implemented independently. 

```suggestion
*   Add `host-tracefs` collector preset.
```

TylerHelmuth (dismissed):

petewall (dismissed):

petewall (approved):

TylerHelmuth (approved):

PR #2839, opened 2026-07-14
Send `presets` attribute to Fleet Management

petewall:
# Description

This sets a new attribute when using remote config, `presets` as a comma-separated list of the collector presets. The intention is that services like Fleet Management can use this attribute to appropriately place configuration based on the collector presets.

## Authorship


charts/k8s-monitoring/tests/platform/grafana-cloud/instrumentation-hub/.rendered/output.yaml
```diff
@@ -162,6 +162,7 @@ data:
         "cluster" = "instrumentation-hub-test-cluster",
         "namespace" = "default",
         "platform" = "kubernetes",
+        "presets" = "root,host-network,host-storage,host-cgroup,host-tracefs,clustered,service-discovery,filesystem-log-reader,daemonset",
         "release" = "k8smon",
         "service-discovery" = "true",
```
petewall: @tpaschalis , if we send the whole preset list, we don't need the `service-discovery` attribute anymore. We could just trigger on something like: `attributes["presets"].split(",").contains("service-discovery")`

TylerHelmuth (approved):

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #2845, opened 2026-07-15
Add the ability to set the ServiceGraph metrics load balancer resolver and timeout

petewall:
# Description

* Create a flag to be able to set the ServiceGraph metrics load balancer resolver. It defaults to `kubernetes`, which matches the current behavior, but can also set:
  * `dns`, which ensures we're using a headless K8s service
  * `static` which ensures we're using a StatefulSet for predictable naming
* Create a flag to expose the timeout for the load balancer resolver.
* Create an examples and tests for service-graph-metrics option in the `otlp` destination with all three resolver options

## Authorship


charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -413,11 +413,27 @@ otelcol.exporter.loadbalancing {{ printf "%s_sampler" (include "helper.alloy_nam
...
+    {{- $collectorName := printf "%s-%s" $.Release.Name (include "helper.k8s_name" (printf "%s-servicegraph" $.destinationName)) | trunc $maxLength | trimSuffix "-" | lower }}
+{{- if eq .processors.serviceGraphMetrics.loadBalancerResolver "kubernetes" }}
     kubernetes {
-      {{- $maxLength := 51 }}{{/* This limit is from the `controller-revision-hash` pod label value*/}}
-      {{- $collectorName := printf "%s-%s" $.Release.Name (include "helper.k8s_name" (printf "%s-servicegraph" $.destinationName)) | trunc $maxLength | trimSuffix "-" | lower }}
       service = "{{ $collectorName }}"
     }
+{{- else if eq .processors.serviceGraphMetrics.loadBalancerResolver "dns" }}
+    dns {
+      hostname = "{{ $collectorName }}.{{ include "helper.namespace" $ }}.svc.cluster.local"
```
TylerHelmuth, line 426: Do we allow customizing the service name or is this a guaranteed name?

petewall: Yeah, it might be possible, but it would be a fairly intentional footgun.

TylerHelmuth: Can we have a validation like "you've set a custom service name but also set `dns` resolution, thats not ok" ?

petewall: let me see if I can get claude to do the right thing.

petewall: Actually never mind.
Alloy doesn't let you override the service name, it's always the same as the alloy full name:
https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/service.yaml#L6

TylerHelmuth: Could we use `"alloy.fullname"` here?

TylerHelmuth (dismissed):

TylerHelmuth (approved):

PR #2846, opened 2026-07-15
feat: add feature to filter otlp logs by pod label/annotation

### LINKED ISSUE #2598: [Feature] podLogsViaOpenTelemetry: allow filters to drop pod logs by annotation/label
## Summary
Add a way to express custom drop rules in the `podLogsViaOpenTelemetry` feature so pods can be excluded from log collection based on their annotations or labels (e.g. drop logs from pods that the OpenTelemetry Operator has auto-instrumented and which already ship logs themselves).

## Motivation
The `podLogsViaLoki` feature exposes `extraDiscoveryRules` (and `extraLogProcessingStages`), which can drop pods by annotation/label regex during discovery. The OpenTelemetry equivalent only supports a single `annotationSelector` + `discoveryMethod` (`all` / `annotation`), where a pod is dropped only when its value is `"false"`, `"no"`, `"skip"`, or `nil`. There is no way to drop based on presence of an unrelated annotation or arbitrary value.

Concrete case: users running the OpenTelemetry Operator with `instrumentation.opentelemetry.io/inject-java` want to skip filelog collection for those pods because the Java agent already exports logs via OTLP. Today the only workaround is to:
1. Extract the annotation into a resource attribute via `podLogsViaOpenTelemetry.annotations`, and
2. Add a `processors.filters.logs.logRecord` rule on every downstream destination.

That works but is per-destination and not discoverable. See discussion in #2597.

## Proposal
Add a `filters` value to the `feature-pod-logs-via-opentelemetry` subchart that appends OTTL log_record statements to the existing `otelcol.processor.filter "pod_logs"` block (`charts/k8s-monitoring/charts/feature-pod-logs-via-opentelemetry/templates/_module.alloy.tpl:118-133`). Rules evaluating to true drop the log, matching the current behavior of that block.

Example usage:
```yaml
podLogsViaOpenTelemetry:
  enabled: true
  annotations:
    inject_java: instrumentation.opentelemetry.io/inject-java
  filters:
    - 'resource.attributes["inject_java"] != nil'
```

## Scope
- `charts/k8s-monitoring/charts/feature-pod-logs-via-opentelemetry/values.yaml` — new `filters: []` key.
- `charts/k8s-monitoring/charts/feature-pod-logs-via-opentelemetry/templates/_module.alloy.tpl` — append rendered entries to the existing filter block.
- `values.schema.json` (subchart and parent) — declare the new property.
- Unit tests under `tests/feature_pod_logs_via_opentelemetry_test.yaml` plus an integration fixture.
- README regeneration via `helm-docs`.

## Alternatives considered
- Reuse `annotationSelector` semantics — rejected; it only drops on `false`/`no`/`skip`/`nil`, not by presence/value of arbitrary annotations.
- Destination-level filter via `processors.filters.logs.logRecord` — works today but must be duplicated per destination and depends on the user remembering to extract the annotation into a resource attribute.

## Related
- #2597

TylerHelmuth:
# Description

Adds a new `filter` section to feature-pod-logs-via-opentelemetry` that allows the user to specify logs to drop based on pod label/annotations.

- Fixes https://github.com/grafana/k8s-monitoring-helm/issues/2598

## Authorship


charts/k8s-monitoring/docs/examples/features/pod-logs-via-opentelemetry/values.yaml
```diff
@@ -24,6 +24,13 @@ podLogsViaOpenTelemetry:
     - production
   namespaceLabels:
     color: color
+  # Surface the OpenTelemetry Operator's Java auto-instrumentation annotation as a resource attribute...
+  annotations:
+    inject_java: instrumentation.opentelemetry.io/inject-java
+  # ...then drop logs from those pods, since the Java agent already exports their logs via OTLP.
+  filters:
+    annotations:
+      inject_java: "true"
```
petewall: Rather than the key being a the resource attribute, can we make this the k8s annotation?

e.g.
```yaml
filters:
  annotations:
    instrumentation.opentelemetry.io/inject-java: true
```

So it doesn't need to be paired with the previous annotations object.

Behind the scenes, we'd still extract that value. maybe as a special attribute like: `filters.annotations.instrumentation.opentelemetry.io/inject-java`

Which we can check, filter, and if the data passes the filter, delete that temporary attribute

petewall (approved):

PR #2862, opened 2026-07-17
feat: add destination routing

### LINKED ISSUE #2849: feat: per-namespace destination routing
**What**

Route telemetry to different destinations based on `k8s.namespace.name`. Today a feature's `destinations:` apply to all data it collects; `namespaces:` only filters collection, not routing. No namespace -> destination mapping exists (confirmed on 4.2.2).

**Why**

Dedicated Grafana Cloud stack per tenant, one tenant per namespace, tenants onboarded continuously. Current options both scale poorly with tenant churn:

- one Helm release per namespace (N releases), or
- N filtered `otlp` destinations in one release (see #2742).

**Proposed shape**

```yaml
destinationRouting:
  attribute: k8s.namespace.name
  table:
    tenant-a: [stack-a]
    tenant-b: [stack-b]
  default: [platform]
```

One release routes each namespace's metrics/logs/traces to its own stack — no per-tenant releases, no hand-authored `custom` processors.

**Relation to #2742**

Overlaps but distinct. #2742 wants dynamic `X-Scope-OrgID` from the namespace attribute on a single destination. Routing solves the manageable-N case of that (one destination per tenantId) but not attribute-derived tenantId. Happy to fold in if maintainers prefer one issue.

Willing to implement — raising for design direction before a PR.

mbaykara:
Fixes #2849

Adds attribute-based destination routing to the k8s-monitoring chart. A new top-level destinationRouting block routes telemetry to different destinations based on a resource attribute (default k8s.namespace.name) from a single Helm release, instead of fanning every feature's data out to all of its destinations.

* Ordered, first-match-wins routes (equals/in/matches), evaluated per signal with an optional per-route signals list
* unmatched data falls through to default.destinations or each feature's own destinations. 
* OTLP destinations route on any resource attribute; Prometheus, Loki, and Pyroscope route on the mapped label. 
* Includes validation, deployment-notes warnings, a docs page, and an example.

outdated, removed this version
> example
> 
> ```yaml
> destinations:
>   stack-a:
>     type: otlp
>     protocol: http
>     url: http://tenant-a.example.com/otlp
>   platform:
>     type: otlp
>     protocol: http
>     url: http://platform.example.com/otlp
> 
> destinationRouting:
>   enabled: true
>   attribute: k8s.namespace.name
>   routes:
>     - match: {equals: tenant-a}
>       destinations: [stack-a]
>   default:
>     destinations: [platform]
> ```
> * Unmatched data goes to default.destinations; when default is omitted it falls through to each feature's own destinations: (current behavior).
>

### DIFF
--- a/charts/k8s-monitoring/AGENTS.md
+++ b/charts/k8s-monitoring/AGENTS.md
@@ -156,6 +156,7 @@ podLogsViaLoki:
 -   `otlp` - Send to any OTLP endpoint (Tempo, Grafana Cloud, etc.)
 -   `prometheus` - Remote write to Prometheus-compatible endpoint
 -   `pyroscope` - Send profiles to Pyroscope
+-   `router` - Route telemetry to other destinations based on an attribute value
 <!-- END DESTINATIONS -->
 
 **Configure a destination:**
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 *   Add the `windows` collector preset, which configures Alloy to run on Windows nodes as a HostProcess container. (@petewall)
+*   Add the `router` destination, which routes telemetry to different destinations based on the value of a resource attribute or label (default `k8s.namespace.name`). Routes are evaluated in order and the first match wins; unmatched data goes to the required `defaultDestinations`. (#2849) (@mbaykara)
 
 ## 4.3.1
 
--- a/charts/k8s-monitoring/Makefile
+++ b/charts/k8s-monitoring/Makefile
@@ -233,6 +233,7 @@ AGENTS.md: AGENTS.md.gotmpl $(FEATURE_CHARTS) $(DESTINATION_VALUES_FILES) $(INTE
 			nop) desc="Drop all telemetry (no storage; for measuring volume before choosing real destinations)";; \
 			otlp) desc="Send to any OTLP endpoint (Tempo, Grafana Cloud, etc.)";; \
 			pyroscope) desc="Send profiles to Pyroscope";; \
+			router) desc="Route telemetry to other destinations based on an attribute value";; \
 			custom) desc="Custom destination configuration";; \
 			*) desc="Send data to $$name";; \
 		esac; \
--- a/charts/k8s-monitoring/destinations/router-values.yaml
+++ b/charts/k8s-monitoring/destinations/router-values.yaml
@@ -1,0 +1,52 @@
+---
+# -- The name for this router destination. A router is not a real telemetry backend: it is a
+# virtual destination that a feature points at (`destinations: [myRouter]`) which then fans
+# telemetry out to real destinations based on an attribute value.
+# @section -- General
+name: ""
+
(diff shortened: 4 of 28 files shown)


petewall:
Rather than mimicking a data processor, does this make sense to implements as a new destination type?

the type could be `router`, and most of the generated components would be the same. The benefit is that none of the internals related to data delivery would need to be changed.

```yaml
destinations:
  tenantRouter:
    type: router
      attribute: k8s.namespace.name
      routes:
        - match: {equals: tenant-a}
          destinations: [stack-a]
      default:
        destinations: [platform]    # Why is this an object with the key `destinations`? Are there other options?

  stack-a:
    type: otlp
    protocol: http
    url: http://tenant-a.example.com/otlp
  platform:
    type: otlp
    protocol: http
    url: http://platform.example.com/otlp


podLogsViaOpenTelemetry:
  enabled: true
  destinations: [tenantRouter]
...
```

I believe all of the template code would only need to go into the router destination template file.

mbaykara:
Thank you, @petewall. I thought about this as well. I’m a bit hesitant to introduce another level under destinations, as I think it might make the UX more complex. But sounds also good to me as touch less internals and your suggestion is more idiomatic

mbaykara:
qq: destinations have one fixed ecosystem, but a router needs to be ecosystem-agnostic per your recommendation. Any preference how to express that in the type helpers?

mbaykara:
I reworked it as a type: router destination and it fits much better and routing lives in destinations: now, features just opt in.
```yaml
destinations:
  tenantRouter:
    type: router
    attribute: k8s.namespace.name
    routes:
      - match: {equals: tenant-a}
        destinations: [stack-a]
    defaultDestinations: [platform]
  stack-a:  
    type: otlp
    protocol: http
    url: http://tenant-a.example.com/otlp
  platform: 
    type: otlp
    protocol: http
    url: http://platform.example.com/otlp

podLogsViaOpenTelemetry:
  enabled: true
  destinations: [tenantRouter]
```
How it works:
each record gets a `selected_destinations` label based on the routes. Routes are checked in order and the first match wins. The record is then sent to the destinations that match. Routing happens on the collection pipeline the data comes in on, so scraped metrics and Loki logs are routed using the mapped label, while OTLP data can be routed on any resource attribute. 
It reuses the gate and target helpers the chart already has, so the way data is actually delivered didn't change.

A few decisions worth calling out:

- `defaultDestinations` is required, and it's a plain list. 
- Data is never dropped without you knowing. Anything that reaches a router either matches a route or goes to defaultDestinations. 
- A feature has to name the router in its destinations: list. Routers are left out of automatic destination selection on purpose, so a feature that doesn't set destinations: can't accidentally send data both to the router and straight to every destination.

petewall (commented):
I like it!

A few things and we'll merge:
1. Can you rebase on top of the "release-4.4" branch? We'll bring it in on that feature release.
2. Can you add a line in the CHANGELOG.

mbaykara:
> I like it!
> 
> A few things and we'll merge:
> 
> 1. Can you rebase on top of the "release-4.4" branch? We'll bring it in on that feature release.
> 2. Can you add a line in the CHANGELOG.

Done!

petewall (approved):

PR #2868, opened 2026-07-17
Replace textlint with docker image

### LINKED ISSUE #2791: Chore: Replace testing and lint clis with docker images
It'd be nice to remove the package.json and yarn.lock files if we can by using docker images, rather than these.

Looks like [markdown_cli](https://github.com/DavidAnson/markdownlint-cli2#install) has a docker image that we could using inside of `make lint-markdown` instead of the npm package.

After they are removed, we can then remove our dependency on node modules, which are only used for testing/linting.

### LINKED ISSUE #2806: Chore: Replace `textlint` with a docker image

petewall:
# Description

Replace the package.json based textlint and subpackages with a docker image/
This also removes our dependency on yarn and npm altogether!

Fixes #2806
Fixes #2791

## Authorship


petewall:
There's no off-the-shelf textlint, so TBD on if I like this build-in-place approach.
Certainly, it's cleaner for us, and running in a container does add sandboxing.

scripts/textlint/Dockerfile
TylerHelmuth, line 1: I think renovate can be told how to watch this file for bumps

TylerHelmuth (approved):

PR #2895, opened 2026-07-24
Fix applicationObservability exemplars.maxPerDataPoint schema

### LINKED ISSUE #2894: [spanmetrics] Exemplars maxPerDataPoint has invalid schema
`applicationObservability.connectors.spanMetrics.exemplars.maxPerDataPoint` should take `int` but only accepts `null`:

https://github.com/grafana/k8s-monitoring-helm/blob/eb9bdc2090014fb1685d322641ab5938769992a4/charts/k8s-monitoring/charts/feature-application-observability/values.schema.json#L74-L84

Also the Type column from the appo11y Readme calls the type "number" instead of "int":

https://github.com/grafana/k8s-monitoring-helm/blob/eb9bdc2090014fb1685d322641ab5938769992a4/charts/k8s-monitoring/charts/feature-application-observability/README.md?plain=1#L118

a-abella:
# Description

Fixes applicationObservability schema restriction disallowing valid/expected values for `connectors.spanMetrics.exemplars.maxPerDataPoint`

Retained the original `null` type in schema as well so the existing default null value remains valid.

Fixes #2894 

## Authorship

### DIFF
--- a/charts/k8s-monitoring/charts/feature-application-observability/values.schema.json
+++ b/charts/k8s-monitoring/charts/feature-application-observability/values.schema.json
@@ -78,7 +78,10 @@
                                     "type": "boolean"
                                 },
                                 "maxPerDataPoint": {
-                                    "type": "null"
+                                    "type": [
+                                        "integer",
+                                        "null"
+                                    ]
                                 }
                             }
                         },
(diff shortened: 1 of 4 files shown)


charts/k8s-monitoring/charts/feature-application-observability/values.schema.json
```diff
@@ -78,7 +78,7 @@
                                     "type": "boolean"
                                 },
                                 "maxPerDataPoint": {
-                                    "type": "null"
+                                    "type": ["integer", "null"]
```
petewall: This change will get overwritten because values.schema.json is generated. Instead add a line in charts/feature-application-observability/schema-mods/types-and-enums.json and it will update this line.

petewall (changes requested):

charts/k8s-monitoring/charts/feature-application-observability/values.schema.json
a-abella: Thank you; I've rebased from main, set the type in schema-mods/types-and-enums.json, and regenerated schema from make

petewall (approved):

petewall:
I hate to delay this more, but our public repos require verified commits. Let me know if you want to try and re-commit with verified commits. Otherwise, I can remake this PR on my own and give you co-authored-by credit.

a-abella:
Amended with a signed commit @petewall

PR #2919, opened 2026-07-29
feat(pod-logs-via-loki): support regex patterns for namespaces

### LINKED ISSUE #2860: Support regex patterns for podLogsViaLoki.namespaces
**Feature request**

Currently, `podLogsViaLoki.namespaces` only accepts an exact list of namespace names:

```yaml
podLogsViaLoki:
  namespaces: []
```

This list is passed directly into Alloy's `discovery.kubernetes` `names` field (see `charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl`), which only supports exact namespace names, not regex/glob patterns.

By contrast, `excludeNamespaces` is applied via a `discovery.relabel` rule with `regex = join("|", ...)`, so it already behaves like a regex/alternation match.

**Request:** allow `namespaces` to support regex patterns (e.g., matching `team-.*` to include all namespaces for a team), similar to how `excludeNamespaces` is applied, rather than requiring an exhaustive exact-match list.

This would likely mean moving namespace inclusion from the `discovery.kubernetes.namespaces.names` field into a `discovery.relabel` `keep` rule with a regex, mirroring the existing exclude logic.

AshwinRamaniPsg:
Move namespace inclusion from discovery.kubernetes's exact-match namespaces.names field to a discovery.relabel keep rule with a regex, mirroring how excludeNamespaces already works. This lets podLogsViaLoki.namespaces match namespaces by pattern (e.g. team-.*) instead of requiring an exhaustive list of exact names.

Fixes #2860

## Authorship

-   [ ] I, a human, wrote this pull request description myself.

Code written with the help of Claude and reviewed by me

### DIFF
--- a/charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl
+++ b/charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl
@@ -21,11 +21,6 @@ discovery.kubernetes "pods" {
     role = "pod"
     field = "spec.nodeName=" + sys.env("HOSTNAME")
   }
-{{- if .Values.namespaces }}
-  namespaces {
-    names = {{ .Values.namespaces | toJson }}
-  }
-{{- end }}
 {{- if $labelSelectors }}
     selectors {
       role = "pod"
@@ -66,6 +61,13 @@ discovery.relabel "filtered_pods" {
     action = "replace"
     target_label = "namespace"
   }
+{{- if .Values.namespaces }}
+  rule {
+    source_labels = ["namespace"]
+    regex = "{{ .Values.namespaces | join "|" }}"
+    action = "keep"
+  }
+{{- end }}
 {{- if .Values.excludeNamespaces }}
   rule {
     source_labels = ["namespace"]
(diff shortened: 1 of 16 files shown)


charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl
```diff
@@ -21,11 +21,6 @@ discovery.kubernetes "pods" {
     role = "pod"
     field = "spec.nodeName=" + sys.env("HOSTNAME")
   }
-{{- if .Values.namespaces }}
```
petewall, line 24: Ideally, I'd still keep this part here _if_ the namespaces are not regex patterns. If we can have them here, we get a lot of performance gains, because the API server gives us a filtered list of pods. If we only do this in discovery.relabel, we get *all* pods and then filter them later. That's a necessary problem with regex namespaces, but with named namespaces, we can avoid it.

TylerHelmuth:
@AshwinRamaniPsg thank you for the issue. We've handled it via https://github.com/grafana/k8s-monitoring-helm/pull/2945

PR #2933, opened 2026-07-31
Add a configuration validator job

petewall:
# Description

This was requested by a user. We had this in v1, but removed in in v2.
Bringing it back, but only optionally.

Enabling this will add a pre-commit hook that will validate the configuration during install time, rather than looking for crashing pods.

## Authorship


petewall:
Just tested this, but at least this test `docs/examples/features/host-metrics/linux-hosts-via-alloy/test` is failing because the prometheus exporter component fails and not the predictable kubernetes failure. We should test more on this.

PR #2941, opened 2026-08-04
update route config to include ecosystem

TylerHelmuth:
# Description

updates the user interface for the router destination to include a designated "ecosystem" and new matching field names. This makes the config more explicit and more extensible in the future. 

## Authorship


charts/k8s-monitoring/schema-mods/destination-router-types-and-enums.json
```diff
@@ -1,9 +1,8 @@
 {
   "definitions": {
     "router-destination": {"properties": {
-      "attribute": {
-        "type": "string",
-        "pattern": "^[a-zA-Z_][a-zA-Z0-9_./-]*$"
+      "ecosystem": {
+        "enum": ["opentelemetry", "prometheus"]
```
petewall: These should match the ecosystems mentioned in the destination's `destination.<type>.ecosystem` template functions
```suggestion
        "enum": ["loki", "otlp", "prometheus", "pyroscope"]
```

For example:
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/templates/destinations/_destination_pyroscope.tpl#L168
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl#L806

petewall (commented):
Matching the ecosystem to those used by the other destinations should mean that we can include a `custom` type destination (which can declare its ecosystem) that can be used in parallel to the others.

petewall (approved):

PR #2945, opened 2026-08-05
Handle regexes in namespaces list

### LINKED ISSUE #2860: Support regex patterns for podLogsViaLoki.namespaces

petewall:
# Description

If a user passes a regex in the list of namespaces, instead of trying to pre-filter inside `discovery.kubernetes` (which only works for absolute namespace names), use the `discovery.relabel` to filter the objects based on the namespace regex.

The following features are enhanced by this change:

* podLogsViaLoki

Fixes #2860

## Authorship


charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl
```diff
@@ -66,6 +70,13 @@ discovery.relabel "filtered_pods" {
     action = "replace"
     target_label = "namespace"
   }
+{{- if eq $namespacesHaveRegex "true" }}
+  rule {  // namespaces contains a regex, so pods cannot be pre-filtered by discovery.kubernetes; keep only matching namespaces here
+    source_labels = ["namespace"]
+    regex = "{{ .Values.namespaces | join "|" }}"
```
TylerHelmuth, line 76: Do we need `()` around each namespace value to make sure the `|` works correctly?

charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_helpers.tpl
```diff
@@ -13,3 +13,19 @@
...
+{{/*
+Detect whether a list of namespaces contains any regular expression, as opposed to only exact namespace
+names. A Kubernetes namespace name must be a valid DNS-1123 label (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`, max 63
+characters), so any entry containing a character outside that alphabet, or exceeding the length limit, can
+only be a regular expression. Outputs "true" if any entry is a regex, and an empty string otherwise.
+*/}}
+{{- define "feature.podLogsViaLoki.namespaces.hasRegex" -}}
+{{- $hasRegex := "false" -}}
+{{- range . -}}
+  {{- if or (not (regexMatch "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$" .)) (gt (len .) 63) -}}
```
TylerHelmuth, line 26: Is there a helm function we can use in a validation to ensure that any in this list that is trying to be a regex really will compile as a regex in alloy?

charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_discovery.alloy.tpl
petewall: Nah, it works fine as it. Plenty of things use this pattern.

charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_helpers.tpl
petewall: Both Helm (via sprig) and Alloy use the same regex syntax. I'm not sure if there's a method for "compiling" the regex without having something to test it against.

TylerHelmuth (approved):

PR #2896, opened 2026-07-24
New Feature: Add the ability to get Windows node metrics from inside of Alloy

petewall:
# Description

The main part of this PR is to be able to use Alloy to get Windows node metrics from inside of Alloy, rather than requiring a separate deployment of the Windows Exporter.

The side-effect of this is that we need to be able to specify collectors for the "subfeatures" of the `hostMetrics` feature. like this:

```yaml
hostMetrics:
  enabled: true
  collectors: my-linux-collector  # The "main" collector
  linuxMetrics:
    enabled: true
  windowsMetrics:
    enabled: true
    collector: my-windows-collector  # Overrides the collector for this subfeature
```
One feature, two different collectors.

Internally, this is done by treating the subfeatures as features of their own, with their own Alloy config module.

Fixes #2661

## Authorship

-   [X] I, a human, wrote this pull request description myself.

PR #2959, opened 2026-08-11
fix: unreleased windows node metrics collection misconfiguration

TylerHelmuth:
# Description

Fixes an issue introduced by https://github.com/grafana/k8s-monitoring-helm/pull/2896 where a copy-pasted resulted in some unneeded configuration. The windows pattern not properly matches the linux pattern. There is already a changelog entry for the original feature that is unreleased so this PR does not need a changelog.

Before this fix the test was failing: https://github.com/grafana/k8s-monitoring-helm/actions/runs/31436584292/job/93629455171

## Authorship

### DIFF
--- a/charts/k8s-monitoring/charts/feature-host-metrics/templates/_windows_hosts.alloy.tpl
+++ b/charts/k8s-monitoring/charts/feature-host-metrics/templates/_windows_hosts.alloy.tpl
@@ -64,9 +64,12 @@ prometheus.relabel "windows_exporter" {
 {{- if .Values.windowsHosts.extraMetricProcessingRules }}
   {{- .Values.windowsHosts.extraMetricProcessingRules | nindent 2 }}
 {{- end }}
-{{- end }}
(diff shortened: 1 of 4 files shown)


jarben (dismissed):
LGTM

jarben:
Only got these non-blocking observations from my agent: 

Pre-existing gap (not this PR's fault): windowsHosts.scheme and windowsHosts.bearerTokenFile are documented but no template consumes them - a Windows scrape behind TLS/auth silently goes out as plain HTTP. Exists in released 4.3.3 too; worth a follow-up issue.

Test gap: no unit test covers source=alloy with extraMetricProcessingRules set - the exact combination this PR fixes. A one-case addition would lock the fix in.

Cosmetic: with all tuning lists empty, the closing brace comment says // prometheus.relabel on what is actually the prometheus.scrape block - misleading comment only, config is valid.

markobachvarovski (approved):

jarben (approved):

PR #2979, opened 2026-08-19
feat: Use prometheus.exporter.static for selfReporting

### LINKED ISSUE #2923: [Self Report] Use the `prometheus.exporter.static` component
Instead of using the prometheus.unix with a static file, let's try using the [prometheus.exporter.static](https://grafana.com/docs/alloy/next/reference/components/prometheus/prometheus.exporter.static/) component.

It is experimental, so I think first it is a spike, then maybe we switch to this when the alloy that's used has the experimental stability level.

petewall:
# Description

When collector is experimental, use prometheus.exporter.static for selfReporting. This avoids using the `prometheus.exporter.unix`, which starts an internal node-exporter and a configmap volume mount, so the pipeline is greatly simplified, and it'll help test this component.

Fixes #2923

## Authorship

### DIFF
--- a/charts/k8s-monitoring/templates/features/_feature_self_reporting.tpl
+++ b/charts/k8s-monitoring/templates/features/_feature_self_reporting.tpl
@@ -42,11 +42,13 @@
 {{- if eq (include "features.selfReporting.enabled" .) "true" }}
   {{- $collectorName := include "features.selfReporting.chooseCollector" . | trim }}
   {{- $collectorValues := (include "collector.alloy.values" (dict "Values" $.Values "Files" $.Files "collectorName" $collectorName) | fromYaml) }}
-  {{- $configMapName := printf "%s-release-info" (include "helper.fullname" .) }}
-  {{- $extraMounts := deepCopy (dig "alloy" "mounts" "extra" list $collectorValues) }}
-  {{- $extraMounts = append $extraMounts (dict "name" "release-info" "mountPath" "/etc/release-info" "readOnly" true) }}
-  {{- $extraVolumes := deepCopy (dig "controller" "volumes" "extra" list $collectorValues) }}
-  {{- $extraVolumes = append $extraVolumes (dict "name" "release-info" "configMap" (dict "name" $configMapName)) }}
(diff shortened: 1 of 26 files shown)


charts/k8s-monitoring/templates/features/_feature_self_reporting.tpl
```diff
@@ -55,12 +57,19 @@ collectors:
...
+
 {{- define "features.selfReporting.validate" }}{{ end }}
+
 {{- define "features.selfReporting.include" }}
 {{- if eq (include "features.selfReporting.enabled" .) "true" }}
-{{- $destinations := include "destinations.get" (dict "destinations" $.Values.destinations "type" "metrics" "ecosystem" "prometheus" "filter" $.Values.selfReporting.destinations) | fromYamlArray -}}
+  {{- $destinations := include "destinations.get" (dict "destinations" $.Values.destinations "type" "metrics" "ecosystem" "prometheus" "filter" $.Values.selfReporting.destinations) | fromYamlArray -}}
+  {{- $collectorName := include "features.selfReporting.chooseCollector" . | trim }}
+  {{- $collectorValues := (include "collector.alloy.values" (dict "Values" $.Values "Files" $.Files "collectorName" $collectorName) | fromYaml) }}
+  {{- $stabilityLevel := dig "alloy" "stabilityLevel" "generally-available" $collectorValues }}
```
TylerHelmuth, line 71: This combo of `collectorName`, `collectorValues`, and the `stabilityLevel` is used a couple times, could we move it to a template?

TylerHelmuth (approved):

charts/k8s-monitoring/templates/features/_feature_self_reporting.tpl
petewall: We define those three lines twice in that file. In one of those places, I'm using `$collectorValues` in other ways aside from getting the stability level. In the other, it's just used for the stability level. So, I think I would make a new template, just to remove one line.

I see the repetition, but I'm just not sure it's worth it in this case.

PR #2983, opened 2026-08-19
Fix inconsistency of how the cluster labels are set in the `pyroscope` and `otlp` destinations

### LINKED ISSUE #2985: Inconsistent handling of cluster labels in destinations
Most destinations do not overwrite an existing cluster label.

`prometheus` and `loki` use their `external_labels` sections, which only set if the label does not already exist.

`otlp` uses this block:

```
    otelcol.processor.transform "tempo" {
      error_mode = "ignore"
    
      trace_statements {
        context = "resource"
        statements = [
          `set(attributes["cluster"], "gateway")`,
          `set(attributes["k8s.cluster.name"], "gateway")`,
           ...
        ]
      }
    }
```
which will overwrite the cluster label always.

`pyroscope` also uses `external_labels`, but it will overwrite the label, unlike `prometheus` and `loki`. This is mentioned in this Alloy issue: https://github.com/grafana/alloy/issues/241

petewall:
# Description

The Pyroscope and OTLP destinations would always overwrite the `cluster` and `k8s.cluster.name` labels, even if those labels/attributes were already set. This is inconsistent with how the `prometheus` and `loki` destinations work, and it causes problems when using a deployment as a gateway.

This change also adds an example around the gateway pattern where different clusters all send their data to a centralized Alloy before delivering to the databases.

Now, this example doesn't really use multiple clusters, but if you squint hard enough, it'll make sense.

Also a few fixes:
* Fix multiple features setting extraPorts, needs to use `mergeOverride` instead of `merge`. Without this, any features that would set a port (like profilesReceiver, and applicationObservability) wouldn't update multiple ports, it would only use the first one.
* Update the Prometheus Metrics Receiver feature to use the more standard port of 9090
* Lowercase the Alloy name, so mixed-case names are permitted.

Fixes #2985

## Authorship

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -8,7 +8,7 @@
 *   Route integration instances to different collectors by setting collector on each instance. (#2616) (@TylerHelmuth)
 *   Route the Profiling sub-features to different collectors by setting collector on each sub-feature. (#2949) (@TylerHelmuth)
 *   When collector is experimental, use the new `prometheus.exporter.static` for selfReporting. (#2923) (@petewall)
-*   The OTLP and Pyroscope destiations no longer overwrite the cluster labels if they're already set. (#2985) (@petewall)
+*   Destinations now have the option to overwrite the cluster labels. Pyroscope and OTLP are enabled by default, maintaining existing behavior. (#2985) (@petewall)
 
 ## 4.4.0
 
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -8,6 +8,7 @@
 *   Route integration instances to different collectors by setting collector on each instance. (#2616) (@TylerHelmuth)
 *   Route the Profiling sub-features to different collectors by setting collector on each sub-feature. (#2949) (@TylerHelmuth)
 *   When collector is experimental, use the new `prometheus.exporter.static` for selfReporting. (#2923) (@petewall)
+*   The OTLP and Pyroscope destiations no longer overwrite the cluster labels if they're already set. (#2985) (@petewall)
 
 ## 4.4.0
 
--- a/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
+++ b/charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
@@ -15,6 +15,24 @@
 {{/* Inputs: . (root object),  destination (string, name of destination), destinationName (name of this destination) */}}
 {{- define "destinations.otlp.alloy" }}
 {{- with .destination }}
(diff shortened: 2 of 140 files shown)


charts/k8s-monitoring/templates/destinations/_destination_otlp.tpl
```diff
@@ -159,13 +159,6 @@ otelcol.processor.transform {{ include "helper.alloy_name" $.destinationName | q
   metric_statements {
     context = "datapoint"
     statements = [
-{{- range $label := .clusterLabels }}
-{{- if $.Values.cluster.nameFrom }}
-      string.format(`set(attributes[%q], %q)`, {{ $label | quote }}, {{ $.Values.cluster.nameFrom }}),
-{{- else }}
-      `set(attributes[{{ $label | quote }}], {{ $.Values.cluster.name | quote }})`,
-{{- end }}
-{{- end }}
```
TylerHelmuth, line 168: Is it purposeful to remove the cluster name from datapoints? If someone sets a value for a cluster label that doesn't automatically index from mimir they wont be able to query it.

petewall: Yeah, I'm going to add that back. I thought it was superfluous, but I think the conversion from otel metrics to prom metrics (when delivering to the OTLP endpoint on a prometheus server) is looking at the metric attribute.

petewall:
This is technically a change in behavior, though I don't anticipate many folks will be affected, since it requires you to forward profiles or OTel data to this chart and actually *want* the cluster label overwritten. I think this change is the much more sensible default, but I could also foresee an issue if people upgrade and they see a sudden change in telemetry data labelling.

I would be happy to add an "Overwrite Cluster Label" option, and defaulting to their current behavior (`false` for Prometheus and Loki, `true` for OTel and Pyroscope). We could then add a note to say that we'll change the defaults in 5.0.

TylerHelmuth:
Ya I think we should provide a way to go back to the previous version unless we want to wait till 5.0

TylerHelmuth (approved):

charts/k8s-monitoring/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 *   Route integration instances to different collectors by setting collector on each instance. (#2616) (@TylerHelmuth)
 *   Route the Profiling sub-features to different collectors by setting collector on each sub-feature. (#2949) (@TylerHelmuth)
 *   When collector is experimental, use the new `prometheus.exporter.static` for selfReporting. (#2923) (@petewall)
+*   The OTLP and Pyroscope destiations no longer overwrite the cluster labels if they're already set. (#2985) (@petewall)
```
TylerHelmuth: Lets add something about how to set it back to the previous behavior incase you were relying on this bug

PR #3004, opened 2026-08-26
Update Alloy CRD deployment instructions

kwentine:
# Description

It seems that the help message indicating how to install the Alloy Operator CRDs is no longer accurate.
The value:
```
alloy-operator:
  crds:
    deployAlloyCRD: true
```
has no effect.

## Authorship

### DIFF
--- a/charts/k8s-monitoring/templates/_crd-validation.tpl
+++ b/charts/k8s-monitoring/templates/_crd-validation.tpl
@@ -2,16 +2,10 @@
 {{- $crdURL := printf "https://github.com/grafana/alloy-operator/releases/download/alloy-operator-%s/collectors.grafana.com_alloy.yaml" (index .Subcharts "alloy-operator").Chart.Version }}
 {{- if .Release.IsInstall }}
   {{- if not (.Capabilities.APIVersions.Has "collectors.grafana.com/v1alpha1/Alloy") }}
-    {{- if not (dig "alloy-operator" "crds" "deployAlloyCRD" true .Values.AsMap) }}
-      {{- $msg := list "" (printf "The %s Helm chart v3.0 requires the Alloy CRD to be deployed." .Chart.Name) }}
-      {{- $msg = append $msg "Please set:" }}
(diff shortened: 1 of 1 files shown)


petewall:
Is the CRD install actually broken?
It actually works if doing a fresh helm install, but it will not install if doing an upgrade. That's why the second block has the instructions to do when upgrading the helm chart.

Do you have an example of installing a fresh install of the chart and the CRD did not get deployed?

kwentine:
Apologies, I made a mistake: since I saw no CRDs in the output of `helm template`, and no logic in this chart using the values mentioned in the PR, I was led to believe that manually installing the Alloy CRDs was the only option left, as hinted in [UPGRADING](https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/UPGRADING.md#alloy-operator) instructions. 

I learned that I should have used:
```
$ helm template k8s-monitoring . \
  --include-crds \
  -f values.debug.yaml
```
and indeed saw the included CRD, using the following `values.debug.yaml`:
```
---
cluster:
  name: test

destinations:
  dev-null:
    type: nop

collectors:
  alloy-metrics:
    presets: [clustered, statefulset]

alloy-operator:
  deploy: true
  crds:
    deployAlloyCRD: true
```

petewall:
No problem at all! CRDs and Helm is a really complicated topic. [This doc](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) says a bit about how it handles CRDs, treating them differently from other templates. IIRC, the major thing is that when you install a CRD, it's available to the whole cluster. If you remove a CRD, it instantly deletes all CRs that are based on it. The danger is if you're on a shared cluster, you may not know (or even have the ability to see) all of the users of a CRD. Helm does the safe thing and says: "We'll install them, but we wont update or delete them."

Some charts get around this by putting them in the templates directory, which works in their case, but only trades one problem for another.

I still appreciate your PR! Thanks for taking the time.

PR #3029, opened 2026-09-10
feat(application-observability): add OTLP HTTP receiver timeout and keep-alive settings

nlamirault:
## Summary

The Application Observability OTLP **HTTP** receiver only exposes `endpoint`, `include_metadata` and `max_request_body_size`, so there is no way to tune the HTTP server's connection lifecycle. The gRPC receiver already exposes `keepalive.server_parameters`, but HTTP has no equivalent.

This surfaces the HTTP server timeout / keep-alive family that `otelcol.receiver.otlp` supports:

- `idleTimeout` → `idle_timeout`
- `readTimeout` → `read_timeout`
- `readHeaderTimeout` → `read_header_timeout`
- `writeTimeout` → `write_timeout`
- `keepAlivesEnabled` → `keep_alives_enabled`

Each attribute is rendered **only when set**, so existing values keep Alloy's built-in defaults and the rendered config is unchanged for current users (verified: existing snapshots unchanged).

https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.receiver.otlp/#http

### DIFF
--- a/charts/k8s-monitoring/CHANGELOG.md
+++ b/charts/k8s-monitoring/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## 4.5.1
 
+*   Add OTLP HTTP receiver timeout and keep-alive settings for the Application Observability feature. (#3029) (@nlamirault)
 *   Fix issue with duplicate configuration block declarations when the same data processor is used by more than one feature on the same collector. (#3014) (@petewall)
 *   Fix incorrect `job` label for nodeLogs feature. (@petewall)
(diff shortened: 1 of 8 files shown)


TylerHelmuth (dismissed):

petewall (changes requested):
This looks good, but can you add an entry to the CHANGELOG.md file?

petewall (dismissed):

TylerHelmuth (approved):

PR #3042, opened 2026-09-14
feat: update PostgreSQL database observability for Alloy 1.19.0

### LINKED ISSUE #3037: Update PostgreSQL Database Observability settings for Alloy 1.19
The PostgreSQL Database Observability integration needs the next Alloy settings sync after #2751, which covered 1.17. The chart exposes four collectors but has no way to feed PostgreSQL server logs into Database Observability.

Alloy added the logs collector in 1.14 (grafana/alloy#5445), error entries with query fingerprints in 1.18 (grafana/alloy#6180), and `user`, `pid` and `sqlstate` fields in 1.19 (grafana/alloy#6770). The fingerprint links an error back to its normalised query.

The source components have to sit inside `declare "postgresql_integration"`, beside the database component. Collector `extraConfig` is appended at root scope and cannot access this module's unexported receiver. That makes this a chart change.

## Proposed configuration

Keep the new settings under `databaseObservability`. The instance-level `logs` key already controls pod-log processing.

For RDS:

```yaml
databaseObservability:
  collectors:
    logs:
      enabled: true
  logSource:
    type: cloudwatch
    cloudwatch:
      region: eu-west-1
      groupName: /aws/rds/instance/example-db/postgresql
```

For a self-hosted instance:

```yaml
databaseObservability:
  collectors:
    logs:
      enabled: true
  logSource:
    type: file
    file:
      paths:
        - /var/log/postgresql/postgresql.log
```

Both sources would forward to `database_observability.postgres.<instance>.logs_receiver` inside the module. Processing would require an enabled setting and a configured source. Existing values must produce byte-identical Alloy configuration, including the current 15s query-sample interval.

The [Alloy 1.19 source](https://github.com/grafana/alloy/blob/v1.19.0/internal/component/database_observability/postgres/component.go) requires `logs.enable_error_logs_processing = true`. Its enable/disable lists are additive, but only cover the four existing configurable collectors. The logs collector starts separately, so adding `"logs"` to `enable_collectors` alone does not enable error processing.

The argument audit also found older chart omissions: `query_samples.enable_pre_classified_wait_events`, `health_check.collect_interval` and the embedded `prometheus_exporter` block. The first two could fit this sync. Embedded exporter support needs a separate decision because it is mutually exclusive with the chart's existing `targets` configuration.

Documentation should cover collector file access, AWS credentials and IAM permissions, RDS log export and the required positional PostgreSQL log prefix. An explicit CloudWatch group name should work without an RDS ARN; deriving a default from the ARN is optional. MySQL stays out of scope.

Does this values structure and scope make sense? - I'll get a draft PR up in a min

rknightion:
# Description

Adds PostgreSQL error-log collection as the next Database Observability settings sync after #2751.

The file and CloudWatch sources sit inside `postgresql_integration`, where they can forward to the instance's `logs_receiver`. Both require `databaseObservability.collectors.logs.enabled` and a configured `databaseObservability.logSource`. The existing pod-log setting keeps its current meaning.

CloudWatch preserves the raw PostgreSQL line through the Loki conversion and stores receiver checkpoints. The examples and generated docs cover persistent storage, file access, AWS credentials and permissions, RDS log export, and the required positional log prefix. The CloudWatch group name is explicit, so it works without an ARN.

The Alloy 1.19 audit also found two older omissions: pre-classified wait events and the health-check interval. Both are now optional settings. The chart keeps its existing 15s query-sample interval. Embedded exporter support needs a separate change because it conflicts with the current `targets` configuration; MySQL is also outside this PR.

Existing values render identically. A three-instance consumer configuration produced the same complete Helm output on main and this branch, and all ten existing PostgreSQL Database Observability snapshots are unchanged.

Fixes #3037. [Maintainer signoff](https://github.com/grafana/k8s-monitoring-helm/issues/3037#issuecomment-5667984314).

## Validation

- `make build`, `make test` and `make lint` pass locally.
- Twenty PostgreSQL tests pass, including assertions for both sources, module scope, opt-in behaviour and invalid source settings.
- Alloy 1.19 validates both source modules and the three-instance CloudWatch consumer render.
- A local mock CloudWatch API with real Alloy confirmed exact raw-body preservation and checkpoint recovery without replay after restart. Bypassing the format hint produced the JSON wrapper the PostgreSQL parser cannot consume.

Live AWS permissions and end-to-end PostgreSQL error telemetry have not been tested.

## Authorship


TylerHelmuth (dismissed):

petewall:
One question I have about the log file reader is where those files are expected to come from in Kubernetes. The path is listed as `/var/log/postgresql/*`, but are those on the Node host? Or in the PostreSQL's pod container volumes?

I'm wondering if there are more generic ways of gathering arbitrary log files from different sources (node file systems, or container volumes).

PR #3044, opened 2026-09-15
feat: add cloudwatch-exporter

schwatvogel:
# Add cloudwatch to k8s-monitoring

## Motivation

As i was using k8s-monitoring quite intensively, and we are not allowed to connect cloudwatch dicetly to our central grafana instance, i needed a low maintenance (in our environment) implementation of alloy.prometheus.cloudwatch.exporter. 
instead of creating an alloy instance manually with just "additional_config" blocks, which will be hard to maintain if you have multiple cross-account configs, i wanted to have the config in the same place, as all other alloy config we have.

## Goal

Beeing able to configure and run multiple cloudwatchexporter instances from values.yml
Cross Account functionality with PodIdentity or other basic aws auth mechanisms

## Longer Term Goals

Provide some sort of prepared Templates/Examples to lessen config Text for basic Tasks (eg "Mohitor cpu mem and network for all ec2 systems tagged with", etc )

## Authorship


petewall:
Thanks for this PR! It feels like a pretty big addition to drop unsolicited. I think we should discuss the opportunity of this addition against the maintainability of it.

Can you describe more about what you're looking for in this feature?

schwatvogel:
i fixed the description, hope that helps a bit. I'm quite happy on helping with the maintainability part of it.

Issue #2123, opened 2025-10-29
issue with dropping the structured_metadata on pogLogs feature


giri0211:
i dont want the `service_instance_id` label from `structured_metadata` coming coming the pod logs feature.
 So I am trying to drop it , but seems there are issues with the chart

I did tried to swtich off structuredMetadata by setting its value to structuredMetadata: {}, but it keeps coming in pod logs
i have tried the second step to label drop the `service_instance_id` from `structuredMetadata` even that did not work
How to get rid of the label `service_instance_id` from pod logs

this is the chart template for pod-logs feature.
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_log_processing.alloy.tpl#L33-L53

Here is my confiurations for podLogs

```yaml
podLogs:
  enabled: true
  # -- Stage blocks to be added to the loki.process component for pod logs.
  # ([docs](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#blocks))
  # This value is templated so that you can refer to other values from this file.
  # @section -- Log Processing
  extraLogProcessingStages: |-
    stage.label_drop {
      values = [
        "service_instance_id",
      ]
    }
  labelsToKeep:
    - region
    - job
    - app
    - namespace
    - pod
    - container
  # recheck if structuredMetadata is really reuired
  # https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/charts/feature-pod-logs/values.yaml
  structuredMetadata: {}
```

Here is the derived logs config map for logs alloy pipeline
```yaml
// Drop the filename label, since it's not really useful in the context of Kubernetes, where we already have cluster,
    // namespace, pod, and container labels. Drop any structured metadata. Also drop the temporary
    // container runtime label as it is no longer needed.
    stage.label_drop {
      values = [
        "filename",
        "tmp_container_runtime",
      ]
    }
    stage.structured_metadata {
      values = {
        "k8s_pod_name" = "k8s_pod_name",
        "pod" = "pod",
        "service_instance_id" = "service_instance_id",
      }
    }
    stage.label_drop {
      values = [
        "service_instance_id",
      ]
    }

    // Only keep the labels that are defined in the `keepLabels` list.
    stage.label_keep {
      values = ["__tenant_id__","region","job","app","namespace","pod","container"]
    }
```

pgroves:
[...]

PR #3052, opened 2026-09-17
Stop indexing service_instance_id as a Loki stream label by default

### LINKED ISSUE #3051: service.instance.id is an unconditional default indexed Loki stream label in podLogsViaLoki -- contradicts docs and v4's own opt-in label model
\`feature-pod-logs-via-loki\`'s discovery template (\`_discovery.alloy.tpl\`) unconditionally sets
\`service_instance_id\` (computed as \`namespace.pod.container\` when no OTel annotation is
present) as a \`target_label\` -- i.e. an indexed Loki stream label -- with no values key to
opt out. Every pod restart/rollout/CronJob run mints a brand-new stream, since the value is
unique per pod instance.

This caused a real incident for us: our staging Loki tenant's active-stream count went from
a flat ~8.7k to ~98k (~11x) within hours of migrating four clusters to v4, tripping the
tenant's max-active-stream limit and dropping ~1k log lines/s via 429s.

**This contradicts Grafana's own documentation.** The official Loki labels guide states
directly: "Because of the potential for high cardinality, \`k8s.pod.name\` and
\`service.instance.id\` are no longer recommended as default labels."
(https://grafana.com/docs/loki/latest/get-started/labels/)

**It also contradicts v4's own stated design philosophy.** Per the "Kubernetes Monitoring
Helm chart v4: Biggest update ever!" blog post, v4 replaced v3's opt-out \`labelsToKeep\` list
with an explicit opt-in model for labels. \`service_instance_id\` bypasses that model entirely --
it's hardcoded and unconditional, the one field with no opt-out.

Related: #2123 reports the same underlying value being hard to remove (and separately, that
\`structuredMetadata: {}\` / \`labels: {}\` don't clear chart defaults due to how Helm coalesces
maps -- posted a working fix there). This issue is about the default itself: should
\`service_instance_id\` be indexed as a label by default at all, given Grafana's own guidance
says it shouldn't be?

**Suggested fix**: either drop the discovery-time \`target_label\` assignment for
\`service_instance_id\` entirely (keep it in \`structuredMetadata\` only, matching how \`pod\`/
\`k8s.pod.name\` are already handled), or gate it behind an explicit opt-in values key, matching
v4's stated label philosophy.

**Chart version**: 4.5.2

dpristovnik-veza:
Fixes #3051.

## Problem

`service_instance_id` is computed as `namespace.pod.container` and is unique per pod
instance, so keeping it as an indexed Loki stream label creates a brand-new stream on
every pod restart, rollout, or Job run -- unbounded over time.

This caused a real incident for us: a Loki tenant's active-stream count went from a flat
~8.7k to ~98k (~11x) within hours of migrating four clusters to v4, tripping the tenant's
max-active-stream limit and dropping ~1k log lines/s via 429s.

Grafana's own [Loki labels documentation](https://grafana.com/docs/loki/latest/get-started/labels/)
already recommends against this: *"Because of the potential for high cardinality,
`k8s.pod.name` and `service.instance.id` are no longer recommended as default labels."*

## Root cause

Both `feature-pod-logs-via-loki` and `feature-pod-logs-via-kubernetes-api` already
compute this value at discovery time and redirect it into `structuredMetadata` by
default (queryable, not indexed) -- but neither ever drops the indexed label copy
afterward, so the value ends up both indexed *and* duplicated into structured metadata.

For comparison, `k8s.pod.name` (the other field these docs call out) is never set as an
indexed label anywhere in either chart -- only reachable via the generic OTel-annotation
`labelmap` rule, an explicit opt-in. `service_instance_id` is the one field that bypasses
that pattern, via its own always-on fallback-synthesis relabel rule.

## Fix

Add a `stage.label_drop` for `service_instance_id` immediately after
`stage.structured_metadata` captures it, in both features' processing templates. This is
gated on the `structuredMetadata` entry for it being present (`hasKey .Values.structuredMetadata
"service.instance.id"`), so a user who explicitly removes that entry (sets it to `null`)
opts back into the previous indexed-label behavior with no other change needed.

No values-schema changes, no new config surface -- this only corrects behavior that
contradicts the chart's own default (`structuredMetadata` already captures the value)
and Grafana's own documentation.

## Testing

- Added a regression test to each feature's `default_test.yaml` asserting the indexed
  label is dropped and the structured-metadata capture is present by default.
- Added a `service_instance_id_label_test.yaml` to each feature asserting the opt-out
  (nulling the `structuredMetadata` entry) restores the old indexed-label behavior.
  Kept as a separate suite file per feature -- some `helm-unittest` runs otherwise share
  the base default-values state across test cases in the same file when one test nulls
  out a nested default key, which made an unrelated default-behavior test in the same
  file order-dependent.
- Updated the 8 pre-existing snapshot files across both features that include this
  block (purely additive -- one new `stage.label_drop` block per snapshot entry, nothing
  else changed).
- `helm lint` / `helm template` against the real chart in both the default and opt-out
  configurations, confirming the discovery-time computation, structured-metadata
  capture, and label-drop ordering all behave as intended.
- `make test` equivalent (`helm unittest`) passes cleanly for both modified features
  (52/52 tests, 42/42 snapshots).

## Related

#2123 reports the same underlying value being hard to remove via `structuredMetadata: {}`
/ `labels: {}` (a separate Helm-map-coalesce gotcha, not fixed here) -- posted a working
workaround there for anyone needing it on an earlier chart version.


TylerHelmuth:
Thanks for opening this pull request. Unfortunately, this contribution is failing to meet our [AI communication guidelines](https://github.com/grafana/k8s-monitoring-helm/blob/main/AGENTS.md#communication).

I'm closing this PR for now. You are very welcome to come back with a PR you have written yourself. Please review the policy linked above before opening a new PR.

Thanks for your understanding.

PR #3053, opened 2026-09-17
Add examples around the different tail sampler resolvers

petewall:
# Description

@ubxt brought in a change to support different resolvers for the tail sampler load balancer. This PR adds examples to demonstrate them.

It also fixes the CHANGELOG and puts it in the right place.

Fixes #

## Authorship


ubxt:
Thank you @petewall

TylerHelmuth (approved):

PR #3057, opened 2026-09-19
Stop indexing service_instance_id as a Loki stream label by default

### LINKED ISSUE #3051: service.instance.id is an unconditional default indexed Loki stream label in podLogsViaLoki -- contradicts docs and v4's own opt-in label model

dpristovnik-veza:
# Description

Fixes #3051

Both `feature-pod-logs-via-loki` and `feature-pod-logs-via-kubernetes-api` index `service_instance_id` as a stream label by default which should by grafana's own docs not be the case. A mass pod-restart event during our fleet migration triggered a sudden spike in active streams, since each new pod gets a new `service_instance_id` value and thus a new indexed stream. Both copy this value into `structuredMetadata` but never drop the indexed label, so it ends up indexed and in metadata. This patch removes it after it was applied to `structuredMetadata` so it does not get indexed but remains in the metadata.

Tested with helm unittest for both features; added test cases for the drop and the opt-out path.

## Authorship


charts/k8s-monitoring/charts/feature-pod-logs-via-loki/templates/_processing.alloy.tpl
```diff
@@ -63,6 +63,18 @@ loki.process "pod_logs" {
   }
 {{- end }}
 
+{{- /* service_instance_id is unique per pod instance (namespace.pod.container), so keeping it
+  as an indexed label creates a brand-new Loki stream on every pod restart/rollout/Job run,
+  unbounded over time -- see https://github.com/grafana/k8s-monitoring-helm/issues/3051. It is
+  already captured above by the default structuredMetadata entry (queryable, not indexed), so
+  drop the indexed copy once captured. Only applies when the structuredMetadata entry for it is
+  present, so removing that entry also opts out of this drop. */ -}}
+{{- if hasKey .Values.structuredMetadata "service.instance.id" }}
```
TylerHelmuth, line 72: The guard on line 58 only checks truthyness so we need to do the same thing here. For example, this values.yaml

```yaml
podLogsViaLoki:
  enabled: true
  structuredMetadata:
    k8s.pod.name: k8s.pod.name
    pod: pod
    service.instance.id: "" 
```
renders

```
stage.structured_metadata {
  values = {
    "k8s_pod_name" = "k8s_pod_name",
    "pod" = "pod",
  }
}
stage.label_drop {
  values = ["service_instance_id"]
}
```

